PR #3, opened 2015-07-27
Maintainer: pkg-go -> Debian Go Packaging Team

aviau:


stapelberg:
http://pkg-go.alioth.debian.org/packaging.html uses pkg-go, and I’m not aware of the team every deciding on a different name than pkg-go. I recall that someone mentioned on IRC that you should use “Debian Go Packaging Team”, but I’m not sure what observation that recommendation was based on.

So, in conclusion, I’m not yet convinced to merge your change — it seems like more places use pkg-go than “Debian Go Packaging Team”. What do other teams use?

aviau:
I don't really have a preference. If you merge this PR, we should update http://pkg-go.alioth.debian.org/packaging.html .

I looked around a little bit and compared with other teams:

- It looks like pkg-java uses: ``Debian Java Maintainers <pkg-java-maintainers@lists.alioth.debian.org>``
- Python modules team uses ``Debian Python Modules Team``. (See [here](http://python-modules.alioth.debian.org/python-modules-policy.html#id2))
- The docker team uses ``Docker Packaging Team <docker-maint@lists.alioth.debian.org>``(see [here](https://sources.debian.net/src/docker.io/1.6.2~dfsg1-2/debian/control/))
- pkg-vim uses ``Debian Vim Maintainers <pkg-vim-maintainers@lists.alioth.debian.org>``(see [here](https://sources.debian.net/src/vim/2:7.4.712-3/debian/control/))
- pkg-TeX uses `` Debian TeX maintainers <debian-tex-maint@lists.debian.org>``(see [here](https://sources.debian.net/src/tex-common/6.02/debian/control/))
- pkg-perl uses ``Debian Perl Group <pkg-perl-maintainers@lists.alioth.debian.org>`` (see [here](https://sources.debian.net/src/dh-make-perl/0.87-1/debian/control/))

stapelberg:
Thanks for checking what other teams use. I think going with a long description, and having that long description be “Debian Go Packaging Team” makes sense.

stapelberg:
I’ve also updated http://pkg-go.alioth.debian.org/packaging.html

aviau:
Thanks :)

PR #4, opened 2015-08-03
Remove "-dev" suffix from "program" binary packages

tianon:
This tracks "debbin" separately from "debsrc" so that the "-dev" suffix can be added only when "-type" is not "program" (because if the program name is "foo", then a binary package name of "foo-dev" is usually not going to be correct).


stapelberg:
Thanks!

tianon:
Glad I could find some little stuff to help with! :smile:

PR #15, opened 2015-09-04
Swap DH_GOPKG for XS-Go-Import-Path

tianon:


stapelberg:
@mika reports that this breaks building for jessie, because the dh-golang version in jessie is too old: http://paste.grml.org/2611/

@tianon Are there any downsides to generating _both_ statements for backwards-compatibility?

PR #16, opened 2015-09-04
Swap hard-coded "unstable" with the slightly more accurate "UNRELEASED" (ready for "dch -r" usage)

tianon:


paultag:
Yeah, totally a fan of this -- makes it a lot easier if someone pushes before dput to NEW

paultag:
(Whoops sorry for the close, anyway has my +1) - @stapelberg ?

stapelberg:
This doesn’t match my work flow so well :|. I never use UNRELEASED (and rarely use dch, in fact), but maybe we should mandate using it in our policy so that expectations are aligned across all of our repositories?

tianon:
I'm pretty +1 (probably obviously, since I made the PR).  It's a common pattern in other teams (debian-python, as one example I've actually worked with/in) especially to help distinguish whether a release has been uploaded or not.

paultag:
Yeah UNRELEASED is pretty standard in most teams - setting release to unstable is normally the job of the sponsor

stapelberg:
I’ve updated https://pkg-go.alioth.debian.org/packaging.html accordingly.

PR #20, opened 2015-09-08
Compress source tarball with xz instead of bzip2

anthonyfok:
Hello,

This suggested pull request might be controversial, as creating .orig.tar.bz2 or .orig.tar.xz is both valid and accepted by Debian, and the choice often comes down to a bit of a personal preference.

That said, there are often small but substantial space saving that we can have by compressing with xz, so I think it would be great for dh-make-golang to use xz for compression by default.  :-)

Or, would you rather make it a configurable option?

### DIFF
--- a/make.go
+++ b/make.go
@@ -95,7 +95,7 @@ func makeUpstreamSourceTarball(gopkg string) (string, string, map[string]bool, s
 	dir := filepath.Dir(gopkg)
 	cmd = exec.Command(
 		"tar",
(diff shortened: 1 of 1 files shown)


tianon:
I'm personally +1 to `.xz` for the compression ratio you mentioned (interesting to note that `.deb`s were switched to use it by default a while back too), and I only didn't make this same PR because I didn't want to paint the shed alone. :innocent:

PR #24, opened 2015-09-11
Built-Using only on programs

aviau:
As discussed [here](http://lists.alioth.debian.org/pipermail/pkg-go-maintainers/Week-of-Mon-20150907/001376.html).

This also changes the field order a little bit. Built-Using will be before Depends.


stapelberg:
I wouldn’t know why field order should matter, so I think this is okay.

PR #29, opened 2015-11-26
Choose the most recent tag even if unannotated

### LINKED ISSUE #22: Consider handling the case where upstream created a lightweight version tag by mistake
Nothing wrong with `dh-make-golang`, but in case issues like russross/blackfriday#196 turn out to be more common in the future, perhaps `dh-make-golang` could try to give a warning or workaround them too?  :-)

anthonyfok:
Print warning if the most recent tag is unannotated (lightweight)
but an older annotated tag exists.

Print info if all tags are unannotated.

Fixes #22

### DIFF
--- a/version.go
+++ b/version.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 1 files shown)


version.go
```diff
@@ -37,17 +38,34 @@ func pkgVersionFromGit(gitdir string) (string, error) {
...
-		lastTag = strings.TrimSpace(string(lastTagBytes)) + "+"
+		lastTag = strings.TrimSpace(string(lastTagBytes))
+
+		// Compare with the most recent annotated tag
+		cmd = exec.Command("git", "describe", "--abbrev=0")
+		cmd.Dir = gitdir
+		if lastAnnotatedTagBytes, err := cmd.Output(); err == nil {
+			lastAnnotatedTag := strings.TrimSpace(string(lastAnnotatedTagBytes))
+			if lastTag != lastAnnotatedTag {
+				log.Printf("WARNING: Lightweight tag %q found, but the most recent annotated tag is %q\n", lastTag, lastAnnotatedTag)
```
stapelberg: Can we make this warning actionable for the user, i.e. suggest what they should do in order to get this resolved?

Also, we should explain what the problem with lightweight tags is.

I’m thinking of something like “Lightweight tags (“git tag v1.0”) are problematic e.g. because they are not considered in “git describe”. Please suggest for upstream to use “git tag v1.0 -as” or similar in order to create an annotated tag instead.”

anthonyfok:
@stapelberg Thank you for the suggestion.

I have added an explanation and suggested action based on your provided text and pushed a new commit to GitHub.  Please let me know if there is anything else that needs to be changed.  Thanks!

Issue #31, opened 2015-12-03
Invalid version number error


hxw:
here is the start of the log:

```
2015/11/30 03:23:39 Downloading "github.com/cihub/seelog/..."
2015/11/30 03:23:42 Determining upstream version number
2015/11/30 03:23:42 Package version is "go1.1+git20151102.84.0f89bb7"
```

Further investigation shows that the author is tagging the "go compiler version" along with "vX.y" tags.
This library has two tags: `go1.1` and `v2.5`  both on the same commit.
Any Idea if it would be possible to pick the 2.5 from the other tag, though it might be easier to be able to specify a version as a command line argument.

stapelberg:
What happens if you specify `-git_revision=v2.5` when running dh-make-golang?

hxw:
I assumed that will checkout that tag rather than the head so will be missing the latest changes, so did not thing to try.  However this is the output:
~~~
2015/12/03 08:55:04 Downloading "github.com/cihub/seelog/..."
2015/12/03 08:55:08 Checking out git revision "v2.5"
2015/12/03 08:55:08 Determining upstream version number
2015/12/03 08:55:08 Package version is "go1.1"
2015/12/03 08:55:08 Determining package type
2015/12/03 08:55:08 Determining dependencies
~~~

the directory afterwards contains:
~~~
golang-github-cihub-seelog/
golang-github-cihub-seelog_go1.1.orig.tar.bz2
itp-golang-github-cihub-seelog.txt
~~~
Also changelog still contains the invalid version:
~~~
golang-github-cihub-seelog (go1.1-1) unstable; urgency=medium
~~~

anthonyfok:
Perhaps adding a command-line option to allow the `--match` flag for `git describe`?

```
foka@debian:~/git/seelog$ git describe 
go1.1-84-g0f89bb7
foka@debian:~/git/seelog$ git describe --match v2.5
v2.5-84-g0f89bb7
```

hxw:
It also accepts a glob
~~~
% git describe --match 'v*'
v2.5-83-g5841159
~~~
could even have a glob that forces a digit: `--match 'v[0-9]*'`

stapelberg:
I like @anthonyfok’s idea. Pull requests which introduce a `-git_describe_match` flag are very welcome.

hxw:
Ok, I'll have a try

hxw:
Seems to work
~~~
[...]

PR #32, opened 2015-12-08
Add a -git_describe_match=<version-glob> option

hxw:
This is to manually override the version detection when the normal
detection returns the wrong tag.  Also to provide a mechanism for
alphabetic prefix stripping where the prefix if no just a lower case 'v'.

examples.
~~~
  -git_describe_match=v2.5
  -git_describe_match='v*'
  -git_describe_match=V22
  -git_describe_match='VrSn*'
~~~

Signed-off-by: Christopher Hall <hsw@ms2.hinet.net>

### DIFF
--- a/make.go
+++ b/make.go
@@ -29,6 +29,10 @@ var (
 		"",
 		"git revision (see gitrevisions(7)) of the specified Go package to check out, defaulting to the default behavior of git clone. Useful in case you do not want to package e.g. current HEAD.")
 
+	gitDescribeMatch = flag.String("git_describe_match",
+		"",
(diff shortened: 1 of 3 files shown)


hxw:
This is to fix #31

stapelberg:
Please remove the other commits from this pull request and open a separate issue so that we can discuss these.

hxw:
The other two were not supposed to be part of this pull, sorry about that.

I'll remove and have another go at just the -git_describe_match and test routine for it.

Thanks for taking time to review.

PR #34, opened 2015-12-15
Add a new -git_describe_match=GLOB option

hxw:
The features are:
* allows for skipping non-version tags
* assumes version tags are 'v' prefixed and are N, N.N  or N.N....N
  (where N represents a decimal number
* the argument is a glob to match with a '*' to represent the wildcard
* example globs: '*' '2.*' '19.*.5'
* the 'v' is optional in the GLOB i.e. this is ok:  `-git_describe_match='v8.*'`
  and has the same meaning as: `-git_describe_match='8.*'`

### DIFF
--- a/make.go
+++ b/make.go
@@ -29,6 +29,10 @@ var (
 		"",
 		"git revision (see gitrevisions(7)) of the specified Go package to check out, defaulting to the default behavior of git clone. Useful in case you do not want to package e.g. current HEAD.")
 
+	gitDescribeMatch = flag.String("git_describe_match",
+		"",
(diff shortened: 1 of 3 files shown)


hxw:
This is a new simplified fix for #31

stapelberg:
> assumes version tags are 'v' prefixed and are N, N.N or N.N....N (where N represents a decimal number

What’s the rationale for this? dh-make-golang shouldn’t make assumptions about upstream version numbers, I think. That way, we avoid some tedious back-and-forth that’s bound to happen as people use dh-make-golang for more and more software.

> the 'v' is optional in the GLOB i.e. this is ok: `-git_describe_match='v8.*'` and has the same meaning as: `-git_describe_match='8.*'`

Why is this necessary? The code would be simpler if we just passed the parameter unmodified to git. Also, version numbers don’t actually *need* to start with a “v”, it’s just a convention that many people use, hence we strip the “v” for convenience. I think stripping it is fine, but requiring it might be a bit too much.

PR #36, opened 2016-01-14
Allow the use of gccgo on architectures without golang-go

anthonyfok:
* Build-Depends on gccgo where golang-go is not available
  i.e., [!amd64 !arm64 !armel !armhf !i386 !ppc64 !ppc64el]

* Remove Depends on golang-go

----

When the debian/control file of a Go program and its dependencies have this change made, this allows the Go program to be built on more platforms, especially powerpc, sparc64, s390x, mips and mipsel, and hopefully more.

For example, see https://buildd.debian.org/status/package.php?p=go-md2man for the current build status of go-md2man.  Of course, my own personal goal is to have [Hugo](https://buildd.debian.org/status/package.php?p=hugo) available on as many platforms as possible.  :wink:

Many thanks!

### DIFF
--- a/make.go
+++ b/make.go
@@ -432,7 +432,7 @@ func writeTemplates(dir, gopkg, debsrc, debbin, debversion string, dependencies 
 	fmt.Fprintf(f, "Maintainer: Debian Go Packaging Team <pkg-go-maintainers@lists.alioth.debian.org>\n")
 	fmt.Fprintf(f, "Uploaders: %s <%s>\n", getDebianName(), getDebianEmail())
 	sort.Strings(dependencies)
-	builddeps := append([]string{"debhelper (>= 9)", "dh-golang", "golang-go"}, dependencies...)
+	builddeps := append([]string{"debhelper (>= 9)", "dh-golang", "golang-any"}, dependencies...)
 	fmt.Fprintf(f, "Build-Depends: %s\n", strings.Join(builddeps, ",\n               "))
 	fmt.Fprintf(f, "Standards-Version: 3.9.8\n")
 	fmt.Fprintf(f, "Homepage: %s\n", websiteForGopkg(gopkg))
@@ -447,7 +447,7 @@ func writeTemplates(dir, gopkg, debsrc, debbin, debversion string, dependencies 
 		fmt.Fprintf(f, "Built-Using: ${misc:Built-Using}\n")
 	} else {
 		fmt.Fprintf(f, "Architecture: all\n")
-		deps = append(append(deps, "golang-go"), dependencies...)
+		deps = append(deps, dependencies...)
 	}
 	fmt.Fprintf(f, "Depends: %s\n", strings.Join(deps, ",\n         "))
 	description, err := getDescriptionForGopkg(gopkg)


stapelberg:
It’s a bit unfortunate that we’d have to maintain the list of architectures if we merged this pull request. Especially since it might change in the future, and then we’d need to update all our packages. I’m not saying this isn’t doable, but I’m not aware of any communication with regards to mass-changes in our packages.

Is there a way with which we can use golang-go or gccgo automatically, depending on availability? E.g. `Depends: golang-go | gccgo`?

anthonyfok:
It is actually _two_ problems, one with `Build-Depends`, and another with `Depends`.  I will talk about them separately.

1. It is indeed a bit unfortunate, but it seems to be the only way it would work with sbuild (buildd.debian.org) because, as we are now all aware, sbuild ignores alternatives in `Build-Depends`, so `Build-Depends: golang-go | gccgo` does not work.

    So, unless we can convince "the powers that be" to change sbuild, the only thing we can do is to change our own packages.

    Or, another way, perhaps, would be to create some kind of a architecture-specific `default-golang` (just a random name that popped into my head) that depends on golang-go or gccgo depending on which architecture it is?  That way, it is only one package to maintain.

2. `Depends: golang-go`, on the other hand, affects not the package itself, in a sense that it can be built, even by sbuild on architectures without golang-go, but it cannot be installed on such architectures, such that the package itself cannot be used by other packages that depend on it, rendering it unusable.

    You will notice that, in my patch, I have simply chosen to remove `Depends: golang-go` altogether, as I have noticed that many older Go packages that were not created with dh-make-golang (such as the ones that @tianon handcrafted) simply did not have `Depends: golang-go`, and they have worked just fine.

    Of course, if you think it is better not to leave it out, we can always use `Depends: golang-go | gccgo`.

As you can see, the troublesome part is with `Build-Depends`, not `Depends`.  :-)

> I’m not saying this isn’t doable, but I’m not aware of any communication with regards to mass-changes in our packages.

Hehe, I was just testing the water with go-md2man because it only depends on two other Go library -dev packages.  But you are right, if we were to follow through, it would eventually result in a mass-change of all related Go packages.  Are you suggesting that we bring this back to pkg-go-maintainers@l.a.d.o for further discussions with everyone?  :-)

anthonyfok:
Oh, I forgot to mention, perhaps the change to architecture-specific `Build-Depends` is unnecessary for "Architecture: all" packages because all of us in the pkg-go team are using platforms where golang-go ***is*** available.  And since they are "Architecture: all" and cover all architectures already, buildd (sbuild) won't try to rebuild them.

So, to limit the impact, we *could* limit the use of architecture-specific Build-Depends, i.e. 

```
Build-Depends: ...
               golang-go [amd64 arm64 armel armhf i386 ppc64 ppc64el],
               gccgo [!amd64 !arm64 !armel !armhf !i386 !ppc64 !ppc64el],
               ...
```

to Go _programs_ only.

anthonyfok:
I have updated the patch to add architecture-specific list to `-type program` only:

> * For programs, Build-Depends on gccgo where golang-go is not available
>  i.e., [!amd64 !arm64 !armel !armhf !i386 !ppc64 !ppc64el]
>
> * For libraries, Depends on "golang-go | gccgo" instead of just "golang-go"

tianon:
For libraries, we only install the source code, so I don't see why `golang-go` would be any more than `Suggests` (if listed at all) -- that's why I've removed it from a bunch of my own pkg-go creations.  Any particular reason we ought to `Depends` on `golang-go` for packages that only include source?

anthonyfok:
+1 on what @tianon said.  :+1:

stapelberg:
Thanks for the thoughtful replies.

> So, unless we can convince "the powers that be" to change sbuild, the only thing we can do is to change our own packages.

Is there a bug report already that we can refer to in a comment? If not, could you file one please? I certainly would be happy if the sbuild issue could be fixed, as we’ve run into that in the past already (with package renames).

> Or, another way, perhaps, would be to create some kind of a architecture-specific default-golang (just a random name that popped into my head) that depends on golang-go or gccgo depending on which architecture it is? That way, it is only one package to maintain.

That sounds reasonable to me, at least as a workaround for the time being. I think a slightly better naming choice might be golang-any, similar to libjson-any-perl or python-anyjson, but I don’t have strong feelings about the name.

@tianon @paultag Do you agree to have `golang-any` in src:golang?

> For libraries, we only install the source code, so I don't see why golang-go would be any more than Suggests (if listed at all)

Actually, I think we can indeed remove the Depends entirely. I was pondering whether having it would be an advantage for people who want to develop against a given library and are not familiar with the ecosystem. However, since the Go libraries in Debian are explicitly _not_ for end-user development, but only for packaging Go programs, this isn’t a valid use-case. So, yes, let’s remove it.

tianon:
I definitely like the `golang-any` solution better than what was proposed in [780355](https://bugs.debian.org/780355). :+1: (Although really interested in @paultag's thoughts on the subject too.)

anthonyfok:
I have just force-pushed a commit removing the Depends on golang-go entirely.  :-)

stapelberg:
@paultag Do you have any objections against a `golang-any` package?

paultag:
Yeah, maintaining a list of arches on every source package seems like it won't scale super well. I don't really want to get into bike shedding a name, so `golang-any` sounds great.

As for the concept of `golang-any`, it's interesting. It matches with my worldview and will give us a single place to pick a golang compiler. This might come in handy once (or rather, if) we need to have two versions of go in the archive at the same time.

I'm with @tianon, sounds better than other ideas so far, anyway.

I say let's go for it.

stapelberg:
Great, it looks like we all agree.

@anthonyfok Could you prepare a patch to add the golang-any package to the golang source package please?

anthonyfok:
> @anthonyfok Could you prepare a patch to add the golang-any package to the golang source package please?

Sorry for the three-month delay.  I finally filed a bug report with patch on the Debian BTS, see https://bugs.debian.org/822746.

However, it turns out that there has been new development: @mwhudson has created src:golang-defaults (similar to srd:gcc-defaults) which provides golang-any and allows the co-installation of multiple versions of Go.  Please see his reply at https://bugs.debian.org/822746 as well as the following proposed changes in the Git repositories on Alioth:

* https://anonscm.debian.org/cgit/pkg-golang/golang.git/log/?h=debian-sid-coinst
* https://anonscm.debian.org/cgit/pkg-golang/golang-defaults.git/

I think @mwhudson's work is a more well-thought-out, future-proof and comprehensive solution.  What do you think?  :-)

anthonyfok:
See also https://bugs.debian.org/818415: "golang: move to per-Go-major version coinstallable packages"

stapelberg:
This looks good to me. However, golang-any is not yet in Debian (only in Ubuntu). Can you ping this PR once golang-any is in Debian? I’ll gladly merge it then.

anthonyfok:
Hi @stapelberg,

> Can you ping this PR once golang-any is in Debian? I’ll gladly merge it then.

Hurray!  golang-any and friends have officially landed in Debian as of today (2016-06-02)!  See https://bugs.debian.org/818415#30, as well as http://tracker.debian.org/pkg/golang-any (which redirects to http://tracker.debian.org/pkg/golang-defaults).

anthonyfok:
Ping again?  @stapelberg

glaubitz:
@anthonyfok Thanks for this working on this!

As a porter, I have actually been wondering how to get Go packages build on more architectures and whether any of the Go packages actually have hard-depends on golang. Very glad to hear that this is being sorted out now, great work!

PS: golang upstream actually wanted to raise the minimum instruction support for ppc64 (Big Endian) to POWER8 to be able to share more code between the ppc64el and ppc64 ports. But I was able to convince upstream to keep the ppc64be port compatible with POWER5 which means we will still be able to use golang on ppc64 even with Go 1.7 :).

anthonyfok:
> @anthonyfok Thanks for this working on this!

@glaubitz Thanks for the compliment, but the real credit goes to @mwhudson for revamping the golang packages to make `golang-any` happen.  :-)

And thank you for convincing the upstream Go team to maintain backward compatibility with POWER5!  ;-)

PR #37, opened 2016-01-29
Allow HTTP(S) proxy environment variables to pass through

anthonyfok:
This allows dh-make-golang to function inside restrictive companies or restrictive countries where Google and/or GitHub servers cannot be access directly.

This also allows `$HOME` to pass through too.  Otherwise, programs like `git` is unable to read `~/.gitconfig` (i.e. `$HOME/.gitconfig`) where the user-defined proxy setting is stored.

Tested to work inside a restrictive country.  :stuck_out_tongue_winking_eye:

Yes, there is a little bit too much code duplication, but I will leave code refactoring in more capable hands, i.e., other developers who are more well-versed in Go than I am.

Cheers,
Anthony

### DIFF
--- a/make.go
+++ b/make.go
@@ -38,6 +38,26 @@ var (
 		"One of \"library\" or \"program\"")
 )
 
+func passthroughEnv() []string {
+	var relevantVariables = []string{
+		"HOME",
+		"PATH",
+		"HTTP_PROXY", "http_proxy",
+		"HTTPS_PROXY", "https_proxy",
+		"ALL_PROXY", "all_proxy",
+		"NO_PROXY", "no_proxy",
+		"GIT_PROXY_COMMAND",
+		"GIT_HTTP_PROXY_AUTHMETHOD",
+	}
+	var result []string
+	for _, variable := range relevantVariables {
+		if value, ok := os.LookupEnv(variable); ok {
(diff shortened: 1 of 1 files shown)


paultag:
Yeah, likely worth refactoring that all out to a single function that's reused, but also dunno if it's worth adding it in this PR. Can't decide on which opinion I have :)

stapelberg:
Thanks for the pull request!

I think refactoring the code like this would be the best move:

```go
func passthroughEnv() []string {
	var relevantVariables = []string{
		"HOME",
		"PATH",
		// …
	}
	var result []string
	for _, variable := range relevantVariables {
		if value, ok := os.LookupEnv(variable); ok {
			result = append(result, fmt.Sprintf("%s=%s", variable, value))
		}
	}
	return result
}
```

Then, when crafting the command, use:

```go
	cmd.Env = append([]string{
		"foo=bar",
	}, passthroughEnv()...)
```

@anthonyfok Can you update the pull request please?

anthonyfok:
@stapelberg, thank you for writing the `passthroughEnv()` function!  It is brilliant and elegant!

I have updated the pull request.  Please check when you have time.

P.S. Sorry for the long delay in my follow-up!

stapelberg:
Thanks!

PR #39, opened 2016-04-12
make.go: bumped Standards-Version, use HTTPS also for Vcs-Git

danstender:
Hi,
I've updated make.go for new standards and HTTPS also for Vcs-Git.
Best,
DS


stapelberg:
Thanks!

PR #44, opened 2016-05-15
also use https url for vcs-git

aviau:


aviau:
wait this was fixed already.

PR #47, opened 2016-09-13
Replace "_" with "-" in program name

sysbot:
This is to ensure it follow Debian Maintainer Guide in regard to
[package](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Source) name

Issue appeared when trying to build a repo with `_`.

```
root@cb38face0702:~# ./bin/dh-make-golang github.com/bitly/oauth2_proxy
2016/09/12 23:07:22 Downloading "github.com/bitly/oauth2_proxy/..."
2016/09/12 23:07:49 Determining upstream version number
2016/09/12 23:07:49 WARNING: Lightweight tag "v2.1" found, but the most recent annotated tag is "v0.1"
2016/09/12 23:07:49     Lightweight tags (created by e.g. "git tag v2.1"
2016/09/12 23:07:49     with no flag) are problematic because, among other
2016/09/12 23:07:49     things, they are ignored by "git describe" by default.
2016/09/12 23:07:49     Please suggest that the upstream replaces the
2016/09/12 23:07:49     lightweight tags with annotated ones.  See
2016/09/12 23:07:49     https://github.com/russross/blackfriday/issues/196
2016/09/12 23:07:49     for details.
2016/09/12 23:07:49
2016/09/12 23:07:49 Package version is "2.1+git20160825.15.a9c55bd"
2016/09/12 23:07:49 Determining package type
2016/09/12 23:07:49 Assuming you are packaging a program (because "github.com/bitly/oauth2_proxy" defines a main package), use -type to override
2016/09/12 23:07:49 Determining dependencies
gbp:error: Couldn't determine upstream package name. Use --interactive.
2016/09/12 23:07:51 Could not create git repository: exit status 1
```

### DIFF
--- a/make.go
+++ b/make.go
@@ -320,12 +320,43 @@ func createGitRepository(debsrc, gopkg, orig string) (string, error) {
 	return dir, nil
 }
 
+// normalize program/source name into Debian standard[1]
+// https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Source
+// Package names (both source and binary, see Package, Section 5.6.7) must
+// consist only of lower case letters (a-z), digits (0-9), plus (+) and minus
+// (-) signs, and periods (.). They must be at least two characters long and
+// must start with an alphanumeric character.
+func normalizeDebianProgramName(str string) string {
+	lowerDigitPlusMinusDot := func(r rune) rune {
+		switch {
+		case r >= 'a' && r <= 'z' || '0' <= r && r <= '9':
+			return r
+		case r >= 'A' && r <= 'Z':
+			return r + ('a' - 'A')
+		case r == '.' || r == '+' || r == '-':
+			return r
+		case r == '_':
+			return '-'
+		}
+		return -1
+	}
+
+	safe := strings.Trim(strings.Map(lowerDigitPlusMinusDot, str), "-")
+	if len(safe) < 2 {
+		return "TODO"
+	}
+
+	return safe
+}
+
 // This follows https://fedoraproject.org/wiki/PackagingDrafts/Go#Package_Names
 func debianNameFromGopkg(gopkg, t string) string {
 	parts := strings.Split(gopkg, "/")
+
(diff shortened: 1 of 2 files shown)


make.go
```diff
@@ -324,7 +324,8 @@ func createGitRepository(debsrc, gopkg, orig string) (string, error) {
 func debianNameFromGopkg(gopkg, t string) string {
 	parts := strings.Split(gopkg, "/")
 	if t == "program" {
-		return parts[len(parts)-1]
+		// XXX: also convert program name to underscore??
```
stapelberg: What does this line mean?

stapelberg:
Thanks for the pull request!

> Package names (both source and binary, see Package, Section 5.6.7) must consist only of lower case letters (a-z), digits (0-9), plus (+) and minus (-) signs, and periods (.). They must be at least two characters long and must start with an alphanumeric character.

Are we confident that underscores are the only thing that needs replacing? Maybe we should extend our approach so that only valid characters are left in the name?

make.go
sysbot: Sorry comments left while debugging. Removing.

stapelberg:
I think using https://golang.org/pkg/strings/#Map might be a better choice than a regular expression here?

I also have some comments about the code as-is, let me know if you’d like me to review that or if I should just review an updated version using Map() :)

stapelberg:
Thanks for your patience, I’m pretty busy these days.

Could you please update the PR to remove the new flag? I’d like to keep dh-make-golang’s interface as minimal as possible. I think Debian tools are too complex in general, so keeping complexity minimal is a very conscious effort I’m making.

stapelberg:
Sorry, should have been clearer: please remove both the flag and the prompting. dh-make-golang already leaves markers in the result that need to be changed, so that is the correct path to require action from the user.

stapelberg:
Thanks!

PR #51, opened 2016-11-20
Use debhelper v10

toddy15:
Hi Michael,

maybe we should Build-Depends on the latest debhelper compatibility level.

Regards,
Tobias


stapelberg:
Thanks!

PR #58, opened 2017-06-23
Add compatibility with go1.2.1

davel:
os.Getenv is available in older versions of Golang. This change allows dh-make-golang to be built using go1.2.1, as distributed in Ubuntu 14.04.

### DIFF
--- a/make.go
+++ b/make.go
@@ -51,7 +51,8 @@ func passthroughEnv() []string {
 	}
 	var result []string
 	for _, variable := range relevantVariables {
-		if value, ok := os.LookupEnv(variable); ok {
+		value := os.Getenv(variable);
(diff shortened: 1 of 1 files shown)


make.go
```diff
@@ -51,7 +51,8 @@ func passthroughEnv() []string {
 	}
 	var result []string
 	for _, variable := range relevantVariables {
-		if value, ok := os.LookupEnv(variable); ok {
+		value := os.Getenv(variable);
+		if (value != "") {
```
stapelberg: more idiomatic:
```go
if value := os.Getenv(variable); value != "" {
```

stapelberg:
Closing this PR due to inactivity.

PR #60, opened 2017-07-03
Add autopkgtest support by default

toddy15:
This is a patch by Martín Ferrari (<tincho@d.o>) which I'm forwarding. I've already applied it in the latest upload of dh-make-golang for Debian.


stapelberg:
Thanks. In the future, it might be prudent to get patches merged before including them in uploads to Debian.

toddy15:
Am 03.07.2017 um 10:54 schrieb Michael Stapelberg:
> Thanks. In the future, it might be prudent to get patches merged before
> including them in uploads to Debian.


You're right, sorry about that.

TheTincho:
Hey all, I did not even remember this repo existed.. Are we not using alioth as the canonical source?

toddy15:
As far as I know, this here on github is the canonical source. However, it would make sense for me to use alioth instead. Michael, what do you think? Time to switch?

stapelberg:
Let’s postpone the switch until alioth was upgraded to software which allows for pull requests and continuous integration (the lack of which, plus networking effects, is the reason why this and other repositories are on GitHub in the first place).

PR #62, opened 2017-07-25
Add a default "Files-Excluded" value for uscan's sake

tianon:
Similar to the `debian/watch` we generate, it seems it'd be useful to also include `Files-Excluded` by default for if/when upstreams switch over to versioned releases which then get consumed via `uscan` instead of via `dh-make-golang`.

Happy to adjust, rebase, amend, discuss, etc. :+1:


stapelberg:
Great point! Thanks!

PR #66, opened 2017-08-15
Try to guess pkg's repo on GitHub

zhsj:
for gopkg.in pkg, it must have a corresponding repo on GitHub, other pkg like pault.ag, he also put related info on reply of https://gopkg/?go-get=1

### DIFF
--- a/make.go
+++ b/make.go
@@ -437,13 +437,6 @@ func getDebianEmail() string {
 	return "TODO"
 }
 
-func websiteForGopkg(gopkg string) string {
-	if strings.HasPrefix(gopkg, "github.com/") {
-		return "https://" + gopkg
-	}
-	return "TODO"
-}
-
 func writeTemplates(dir, gopkg, debsrc, debbin, debversion string, dependencies []string) error {
 	if err := os.Mkdir(filepath.Join(dir, "debian"), 0755); err != nil {
 		return err
@@ -489,7 +482,7 @@ func writeTemplates(dir, gopkg, debsrc, debbin, debversion string, dependencies 
 	builddeps := append([]string{"debhelper (>= 10)", "dh-golang", "golang-any"}, dependencies...)
 	fmt.Fprintf(f, "Build-Depends: %s\n", strings.Join(builddeps, ",\n               "))
 	fmt.Fprintf(f, "Standards-Version: 4.0.0\n")
-	fmt.Fprintf(f, "Homepage: %s\n", websiteForGopkg(gopkg))
+	fmt.Fprintf(f, "Homepage: %s\n", getHomepageForGopkg(gopkg))
 	fmt.Fprintf(f, "Vcs-Browser: https://anonscm.debian.org/cgit/pkg-go/packages/%s.git\n", debsrc)
 	fmt.Fprintf(f, "Vcs-Git: https://anonscm.debian.org/git/pkg-go/packages/%s.git\n", debsrc)
 	fmt.Fprintf(f, "XS-Go-Import-Path: %s\n", gopkg)
@@ -536,7 +529,7 @@ func writeTemplates(dir, gopkg, debsrc, debbin, debversion string, dependencies 
 	}
 	fmt.Fprintf(f, "Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/\n")
 	fmt.Fprintf(f, "Upstream-Name: %s\n", filepath.Base(gopkg))
-	fmt.Fprintf(f, "Source: %s\n", websiteForGopkg(gopkg))
+	fmt.Fprintf(f, "Source: %s\n", getHomepageForGopkg(gopkg))
 	fmt.Fprintf(f, "Files-Excluded: vendor Godeps/_workspace\n")
 	fmt.Fprintf(f, "\n")
 	fmt.Fprintf(f, "Files: *\n")
@@ -636,7 +629,7 @@ func writeITP(gopkg, debsrc, debversion string) (string, error) {
 	fmt.Fprintf(f, "* Package name    : %s\n", debsrc)
 	fmt.Fprintf(f, "  Version         : %s\n", debversion)
 	fmt.Fprintf(f, "  Upstream Author : %s\n", author)
-	fmt.Fprintf(f, "* URL             : %s\n", websiteForGopkg(gopkg))
+	fmt.Fprintf(f, "* URL             : %s\n", getHomepageForGopkg(gopkg))
 	fmt.Fprintf(f, "* License         : %s\n", license)
 	fmt.Fprintf(f, "  Programming Lang: Go\n")
 	fmt.Fprintf(f, "  Description     : %s\n", description)
--- a/metadata.go
+++ b/metadata.go
@@ -4,8 +4,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
+
+	"golang.org/x/net/html"
 )
 
 type licensesReply struct {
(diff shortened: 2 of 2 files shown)


make.go
```diff
@@ -489,7 +482,7 @@ func writeTemplates(dir, gopkg, debsrc, debbin, debversion string, dependencies
 	builddeps := append([]string{"debhelper (>= 10)", "dh-golang", "golang-any"}, dependencies...)
 	fmt.Fprintf(f, "Build-Depends: %s\n", strings.Join(builddeps, ",\n               "))
 	fmt.Fprintf(f, "Standards-Version: 4.0.0\n")
-	fmt.Fprintf(f, "Homepage: %s\n", websiteForGopkg(gopkg))
+	fmt.Fprintf(f, "Homepage: %s\n", getHomepageForGopkg(gopkg))
```
stapelberg: Why did you rename this function? Can we keep the old name to make this change minimal?

metadata.go
```diff
@@ -80,12 +82,36 @@ var debianLicenseText = map[string]string{
 `,
 }
 
+var githubRegexp = regexp.MustCompile(`github\.com/([\w._-]+/[\w._-]+)`)
```
stapelberg: Why not just `github.com/([^/]+/[^/]+)`?

metadata.go
```diff
@@ -80,12 +82,36 @@ var debianLicenseText = map[string]string{
...
+
+// regex match github.com/([^/]+/[^/]+) on reply of https://gopkg/?go-get=1
```
stapelberg: Remove the comment, it just re-iterates implementation details. Having one place to change is better than having to change two places :)

metadata.go
```diff
@@ -80,12 +82,36 @@ var debianLicenseText = map[string]string{
...
+func findGitHubRepo(gopkg string) (string, error) {
+	if strings.HasPrefix(gopkg, "github.com/") {
+		return gopkg[len("github.com/"):], nil
```
stapelberg: slightly cleaner: `return strings.TrimPrefix(gopkg, "github.com/"), nil`

metadata.go
```diff
@@ -80,12 +82,36 @@ var debianLicenseText = map[string]string{
...
+	resp, err := http.Get("https://" + gopkg + "?go-get=1")
+	if err != nil {
+		return "", err
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
+	if err != nil {
+		return "", err
+	}
+	repoMatch := githubRegexp.FindSubmatch(body)
```
stapelberg: Instead of just applying a regexp, I think it’d be more robust/cleaner to actually extract the meta tags. See https://godoc.org/golang.org/x/net/html#example-Parse for an example of how to do that.

make.go
zhsj: I want to move this func to metadata.go since it's related and uses findGitHubRepo func. other func in metadata.go has getXxxForGopkg schema.

metadata.go
zhsj: After I first committed this, and try to test with https://pault.ag/go/config?go-get=1 it matches `https://github.com/paultag/go-config"`

metadata.go
zhsj: well, like gopkg.in,
```
<meta name="go-import" content="gopkg.in/go-playground/validator.v8 git https://gopkg.in/go-playground/validator.v8">
<meta name="go-source" content="gopkg.in/go-playground/validator.v8 _ https://github.com/go-playground/validator/tree/v8.18.2{/dir} https://github.com/go-playground/validator/blob/v8.18.2{/dir}/{file}#L{line}">
```
it only show the github repo in go-source meta. And we still need to do a regexp match for the content of go-source.

But go-source is not always shown. like pault.ag,

```
                <meta charset="utf-8">
                <title>pault.ag/go/config</title>
                <meta name="go-import" content="pault.ag/go/config git https://github.com/paultag/go-config">
        </head>
```
paultag only uses go-import :)

IMHO, regexp is much simpler.

make.go
stapelberg: Okay, that’s fair.

metadata.go
stapelberg: I don’t doubt that it works with the cases you’ve tested, but my suggestion is clearer, simpler and more reliable: after all, slashes are the path separators in URLs, so that’s what we should work with.

metadata.go
stapelberg: regexp indeed is simpler, but my concern is not about simplicity, but reliability. It’s probably fine to start with the regexp approach, but please add a TODO to actually parse go-import and go-source.

metadata.go
zhsj: Before we parse html and match this only on meta content, we can't just use `[^/]`. Too much corner cases for regex match on plain html :(

metadata.go
zhsj: Added a TODO

metadata.go
stapelberg: Can you add a comment which describes this issue and contains such an edge case please?

zhsj: Added

metadata.go
```diff
@@ -80,12 +82,38 @@ var debianLicenseText = map[string]string{
 `,
 }
 
+// We don't use github\.com/([^/]+/[^/]+), since there's something like
+// <meta ... git https://github.com/owner/repo"> and this regex will fail.
```
stapelberg: Does `github\.com/([^/]+/[^/]+)\b` do the trick, then?

zhsj: hmm, failed with `github.com/xx/xx-`
seems I need to write some tests for it.

zhsj: Added tests and removed the old comment(I think it's duplicated)

metadata_test.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	"testing"
+)
+
+func TestGithubRegexp(t *testing.T) {
+	cases := map[string]string{
+		`git https://github.com/paultag/go-config">`: "paultag/go-config",
+		`https://github.com/owner/repo->`:            "owner/repo-",
```
stapelberg: Where did you see this in practice? Do you have a URL where I can find such a tag?

metadata_test.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	"testing"
+)
+
+func TestGithubRegexp(t *testing.T) {
```
stapelberg: Can you mark this test as safe for parallel execution using `t.Parallel()` please?

metadata.go
stapelberg: How about `github\.com/([^/]+/[^/>" ]+)`? To elaborate on why I’m being so difficult about this regexp: in general, I’d like this program to be permissive by default (using a blacklist of forbidden characters), instead of restrictive by default (using a whitelist of allowed characters). That seems more robust to me.

If my latest suggestion doesn’t work either, we’ll stick with yours and I’ll implement the HTML parsing afterwards :).

metadata_test.go
zhsj: did you mean `https://github.com/owner/repo->`? Actually I haven't seen someone using `-` in his repo name. But it's legal for GitHub, just created a test repo https://github.com/zhsj/test-

stapelberg: Okay. How about the missing `"`, though? Did you observe that in the wild?

zhsj: (don't know why I can't reply the last commet, so reply here)
seems if we want to relax the pattern, more situations needed to consider.
I'm fine to give up using restricted regex pattern. Let's use HTML parser instead, which is not a tough task actually. I will implement this later :)

metadata.go
```diff
@@ -80,12 +83,69 @@ var debianLicenseText = map[string]string{
...
+	if strings.HasPrefix(gopkg, "github.com/") {
+		repo = strings.TrimPrefix(gopkg, "github.com/")
+		return
+	}
+	resp, err := http.Get("https://" + gopkg + "?go-get=1")
+	if err != nil {
+		return
+	}
+	defer resp.Body.Close()
+	z := html.NewTokenizer(resp.Body)
```
stapelberg: Are you intentionally not using the Parse method? https://godoc.org/golang.org/x/net/html#example-Parse

It usually results in shorter and easier to follow code :)

metadata.go
```diff
@@ -80,12 +83,69 @@ var debianLicenseText = map[string]string{
 `,
 }
 
+var githubRegexp = regexp.MustCompile(`github\.com/([^/]+/[^/]+)`)
+
+func findGitHubRepo(gopkg string) (repo string, err error) {
+	if strings.HasPrefix(gopkg, "github.com/") {
+		repo = strings.TrimPrefix(gopkg, "github.com/")
+		return
```
stapelberg: Here and elsewhere, please don’t use naked returns:

> Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values. Corollary: it's not worth it to name result parameters just because it enables you to use naked returns. Clarity of docs is always more important than saving a line or two in your function.

(from https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters)

metadata.go
```diff
@@ -80,12 +83,69 @@ var debianLicenseText = map[string]string{
...
+	defer resp.Body.Close()
+	z := html.NewTokenizer(resp.Body)
+	for {
+		tt := z.Next()
+		if tt == html.ErrorToken {
+			err = fmt.Errorf("%q is not on GitHub", gopkg)
+			return
+		}
+		token := z.Token()
+		fmt.Println(token.String())
```
stapelberg: Debugging left-over? Please remove.

metadata.go
zhsj: At first, I just randomly picked one of the methods.
I tried to use html.Parse after your suggestion, but seems it's not shorter. The most code is to parse meta content manually.
And I find if to walk through a node link, more code is needed to jump out when first github regexp matches.
So I'd like to use html.NewTokenizer.

metadata.go
zhsj: fixed(maybe I have some wrong impressions for named result parameters..)

stapelberg:
Thanks again!

anthonyfok:
This pull request fixes [Debian Bug \#872015](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=872015)

PR #67, opened 2017-08-18
Upgrade Standards-Version to 4.0.1

zhsj:
change priority to optional

from upgrading checklist:

 The extra priority has been deprecated and should be treated as
 equivalent to optional. All extra priorities should be changed to
 optional.

Signed-off-by: Shengjing Zhu <i@zhsj.me>


stapelberg:
Thanks!

PR #68, opened 2017-09-25
Upgrade Standards-Version to 4.1.0

ctrochalakis:
Hello! Just a bump to the `Standards-Version`.

See https://www.debian.org/doc/debian-policy/index.html#version-4-1-0


stapelberg:
Thanks!

PR #70, opened 2017-10-12
Compress repacked tarballs with xz and append +dfsg1 on repacking

rbalint:


tianon:
I'm not the maintainer, but could we please use `~` instead of `+` by default (so that `+` can be used as an escape hatch)?  Also, I think `ds1` would probably be more accurate (since we're not repacking due to non-freeness, just due to code copies policy).

stapelberg:
Thanks for the pull request.

Aside from tianon’s comment (which I agree with), I’m not convinced that it is a good trade-off to lengthen all our version numbers for saving a few bytes of disk.

Is there any Debian project-wide guidance/advice from which this suggestion follows, or is this your personal preference for xz files?

anthonyfok:
Fixed in commit 554ef2066a77c610cd83b532ef9778db9fd37021 which ended up in v0.3.3 release.
commit 554ef2066a77c610cd83b532ef9778db9fd37021 differs from this PR in the following ways:

* repack only if necessary, i.e. if Files-Excluded field exists in debian/copyright
* repacksuffix: +ds1
* compression method not specified, though that usually means xz in practice.

Issue #76, opened 2017-12-28
Allow for searching packages


aviau:
Hello,

Would you merge a feature that allows for searching existing packages by gopath?

For example:
```shell
$ dh-make-golang search tomb
golang-gopkg-tomb.v1-dev: gopkg.in/tomb.v1
golang-gopkg-tomb.v2-dev: gopkg.in/tomb.v2
```

I know that dh-make-golang warns when packaging stuff that is already packaged but I would find this very useful for searching what is already available in Debian.

Example use case: I imported a new upstream version for my existing package and it won't build anymore because of a new dependency. I run ``dh-make-golang search <keyword>`` to quickly find the debian binary package name and add it to build-dependencies.

PR #77, opened 2017-12-28
create search command

aviau:
Closes: #76

Note that this PR does **not** include #75. Feel free to merge them separately, in any order. I will also gladly rebase one on the other if needed.

see my commit message for details:

````
Example usage:
```
    $ dh-make-golang search debian
    golang-godebiancontrol-dev: github.com/stapelberg/godebiancontrol
    golang-pault-go-debian-dev: pault.ag/go/debian
```

Example use case: I imported a new upstream version for my existing
package and it won't build anymore because of a new dependency. I run
``dh-make-golang search <keyword>`` to quickly find the Debian binary
package name and add it to the build dependencies.
````

I know that dh-make-golang warns when packaging stuff that is already packaged but I would find this very useful for searching what is already available in Debian.

### DIFF
--- a/search.go
+++ b/search.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"regexp"
+	"strings"
+)
+
+const (
+	golangBinariesURL = "https://api.ftp-master.debian.org/binary/by_metadata/Go-Import-Path"
+)
+
+func getGolangBinaries() (map[string]string, error) {
+	golangBinaries := make(map[string]string)
+
+	resp, err := http.Get(golangBinariesURL)
+	if err != nil {
+		return nil, fmt.Errorf("getting %q: %v", golangBinariesURL)
+	}
+	if got, want := resp.StatusCode, http.StatusOK; got != want {
+		return nil, fmt.Errorf("unexpected HTTP status code: got %d, want %d", got, want)
+	}
+	var pkgs []struct {
+		Binary     string `json:"binary"`
+		ImportPath string `json:"metadata_value"`
+		Source     string `json:"source"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&pkgs); err != nil {
+		return nil, err
+	}
+	for _, pkg := range pkgs {
+		if !strings.HasSuffix(pkg.Binary, "-dev") {
+			continue // skip -dbgsym packages etc.
+		}
+		golangBinaries[pkg.ImportPath] = pkg.Binary
+	}
+	return golangBinaries, nil
+}
+
+func execSearch(args []string) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: search.go [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+		log.Fatal(err)
+	}
+
+	if fs.NArg() != 1 {
+		fmt.Fprintf(os.Stderr, "Usage: %s search <pattern>\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "Example: %s search debian\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	pattern, err := regexp.CompilePOSIX(fs.Arg(0))
```
stapelberg: Why POSIX syntax instead of Go’s default regexp syntax?

search.go
```diff
@@ -0,0 +1,75 @@
...
+func execSearch(args []string) {
+	fs := flag.NewFlagSet("search", flag.ExitOnError)
+
+	err := fs.Parse(args)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if fs.NArg() != 1 {
+		fmt.Fprintf(os.Stderr, "Usage: %s search <pattern>\n", os.Args[0])
```
stapelberg: Please document which regexp syntax `pattern` uses.

aviau: Okay, done.

search.go
aviau: Oh, yeah it makes sense to use the default.

aviau:
I have applied your recommended changes.

search.go
```diff
@@ -0,0 +1,76 @@
...
+		fmt.Fprintf(os.Stderr, "Uses Go's default regexp syntax\n")
```
stapelberg: Let’s add https://golang.org/pkg/regexp/syntax/ to the message to help people who are not yet familiar with the syntax.

search.go
```diff
@@ -0,0 +1,76 @@
...
+		fmt.Fprintf(os.Stderr, "Example: %s search debian\n", os.Args[0])
```
stapelberg: Could we make the example use a common regexp so that people can easily adapt the example without having to consult the whole documentation? How about “debi.*” for example?

search.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	golangBinaries, err := getGolangBinaries()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	for importPath, binary := range golangBinaries {
+		result := pattern.FindAllString(importPath, -1)
+		if len(result) > 0 {
```
stapelberg: Simpler:
```go
if pattern.MatchString(importPath) {
```

aviau: Done!

search.go
aviau: Good idea, done!

search.go
aviau: Alright! I have switched to your example!

aviau:
I have applied your recommended changes!

PR #82, opened 2018-01-10
add instructions to create repos on salsa

aviau:


stapelberg:
I don’t think this is a good idea. Has there been a discussion about how we’ll approach our move to salsa? Is salsa out of beta yet?

I’d prefer it if we could have one and only one place to store repos, not two.

aviau:
I have modified the instructions to use the "create-salsa-project" command that will be introduced in #82 . 

We can merge this after we officially move to salsa.

make.go
```diff
@@ -821,10 +821,10 @@ func execMake(args []string) {
 	log.Printf("    git add debian && git commit -a -m 'Initial packaging'\n")
 	log.Printf("    gbp buildpackage --git-pbuilder\n")
 	log.Printf("\n")
-	log.Printf("To create the packaging git repository on alioth, use:\n")
-	log.Printf("    ssh git.debian.org \"/git/pkg-go/setup-repository %s 'Packaging for %s'\"\n", debsrc, debsrc)
+	log.Printf("To create the packaging git repository on salsa, use:\n")
+	log.Printf("    SALSA_TOKEN=\"<token here>\" dh-make-golang create-salsa-project %s", debsrc)
```
stapelberg: The SALSA_TOKEN can be removed now that create-salsa-project is done server-side.

make.go
```diff
@@ -821,10 +821,10 @@ func execMake(args []string) {
...
 	log.Printf("\n")
 	log.Printf("Once you are happy with your packaging, push it to alioth using:\n")
-	log.Printf("    git remote set-url origin git+ssh://git.debian.org/git/pkg-go/packages/%s.git\n", debsrc)
+	log.Printf("    git remote set-url origin git@salsa.debian.org:pkg-go-team/packages/%s.git\n", debsrc)
```
stapelberg: s/pkg-go-team/go-team/

make.go
aviau: done!

make.go
aviau: done!

stapelberg:
Thanks!

PR #83, opened 2018-01-26
add create-salsa-project command

aviau:
I think that this can be merged right now. It allows creating repositories on salsa but does not recommend it yet.

This way, it will be available to everyone as soon as we migrate to salsa.

Then, we can merge #82 when we officially migrate

This command:
 - creates a repository in pkg-go-team/packages
 - adds a webhook to tag bugs as pending

### DIFF
--- a/create_salsa_project.go
+++ b/create_salsa_project.go
@@ -1,0 +1,134 @@
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"strconv"
+)
+
+const (
+	salsaGroupID = 2638
+	salsaApiUrl  = "https://salsa.debian.org/api/v4"
+)
+
+type gitlabProject struct {
+	ID            int    `json:"id"`
+	Name          string `json:"name"`
+	HttpUrlToRepo string `json:"http_url_to_repo"`
+}
+
+func execCreateSalsaProject(args []string) {
+	projectName := mustGetProjectName(args)
+	token := mustGetSalsaToken()
+
+	gitlabProject, err := createSalsaProject(projectName, token)
+	if err != nil {
+		log.Fatalf("Could not create %s on salsa: %s\n", projectName, err)
+	}
+
+	fmt.Printf("Project %s was created on salsa.debian.org: %s\n",
+		gitlabProject.Name,
+		gitlabProject.HttpUrlToRepo)
+
+	if err := createSalsaWebhook(gitlabProject.ID,
+		"https://webhook.salsa.debian.org/tagpending/"+gitlabProject.Name,
+		token); err != nil {
+		log.Fatalf("Could not create webhook on salsa project: %s\n", err)
+	}
+
+}
+
+func postFormToSalsaApi(path string, data url.Values, token string) (*http.Response, error) {
+	postUrl := salsaApiUrl + path
+	data.Add("private_token", token)
+	return http.PostForm(postUrl, data)
+}
+
+func createSalsaWebhook(projectId int, webhookUrl, token string) error {
+	postPath := "/projects/" + strconv.Itoa(projectId) + "/hooks"
+
+	response, err := postFormToSalsaApi(postPath,
+		url.Values{
+			"url":         {webhookUrl},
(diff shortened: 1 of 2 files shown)


create_salsa_project.go
```diff
@@ -0,0 +1,166 @@
...
+	Name          string `json:"name"`
+	HttpUrlToRepo string `json:"http_url_to_repo"`
+}
+
+func execCreateSalsaProject(args []string) {
+	projectName := getProjectNameOrExit(args)
+	token := getSalsaTokenOrExit()
+
+	gitlabProject, err := createSalsaProject(projectName, token)
+
```
stapelberg: here and elsewhere: remove newline before error handling instructions, they logically belong to the actual instructions

create_salsa_project.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Could not create %s on salsa: %s\n", projectName, err)
```
stapelberg: Here and elsewhere, instead of fmt.Fprintf+os.Exit, use log.Fatalf.

create_salsa_project.go
```diff
@@ -0,0 +1,166 @@
...
+		os.Exit(1)
+	}
+
+	fmt.Printf(
```
stapelberg: For consistency, please format as:

```go
fmt.Printf("Project …",
  foo)
```

create_salsa_project.go
```diff
@@ -0,0 +1,166 @@
...
+		"Project %s was created on salsa.debian.org: %s\n",
+		gitlabProject.Name,
+		gitlabProject.HttpUrlToRepo,
+	)
+
+	err = createSalsaWebhook(
```
stapelberg: More idiomatic:

```go
if err := createSalsaWebhook(…); err != nil {
  log.Fatalf(…)
}
```

create_salsa_project.go
```diff
@@ -0,0 +1,166 @@
...
+		},
+		token,
+	)
+
+	if err != nil {
+		return err
+	}
+
+	if response.StatusCode != http.StatusCreated {
+		responseData, err := ioutil.ReadAll(response.Body)
```
stapelberg: It is more common to ignore the error when reading response bodies. The caller will still want to know the actual error (http status mismatch in this case), not just (e.g.) io.EOF

create_salsa_project.go
```diff
@@ -0,0 +1,166 @@
...
+		if err != nil {
+			return err
+		}
+
+		return errors.New(
```
stapelberg: Use fmt.Errorf instead of errors.New+fmt.Sprintf

create_salsa_project.go
```diff
@@ -0,0 +1,166 @@
...
+			fmt.Sprintf(
+				"http status %d: %s",
+				response.StatusCode,
+				responseData,
+			),
+		)
+	}
+
+	return err
```
stapelberg: nit: `return nil`

create_salsa_project.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func getProjectNameOrExit(args []string) string {
```
stapelberg: here and below: it is more idiomatic to use the `must` prefix than the `OrExit` suffix, i.e. make this `mustGetProjectName` please

create_salsa_project.go
aviau: Done!

create_salsa_project.go
aviau: Done.

create_salsa_project.go
aviau: Done!

create_salsa_project.go
aviau: Done.

create_salsa_project.go
aviau: In this case, errors contains informations such as "repo already exists". Its returned in json but I don't think its worth parsing.

create_salsa_project.go
aviau: Done!

create_salsa_project.go
aviau: done

aviau:
Thank you for your feedback. I have applied your recommendations.

PR #84, opened 2018-01-29
protect salsa project branches

aviau:
This will disable force pushes on the following branches:

- master
 - debian/*
- upstream
- pristine-tar

This covers the current and the future repository format that will be used by the team.

Currently, dh-make-golang creates repositories with the pristine-tar branch but in the future  we will switch to debian/*. We might as well protect the branches already.

### DIFF
--- a/create_salsa_project.go
+++ b/create_salsa_project.go
@@ -43,12 +43,41 @@ func execCreateSalsaProject(args []string) {
 		log.Fatalf("Could not create webhook on salsa project: %s\n", err)
 	}
 
+	for _, branch := range []string{"master", "debian/*", "upstream", "upstream/*", "pristine-tar"} {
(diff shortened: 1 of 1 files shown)


create_salsa_project.go
```diff
@@ -43,6 +43,12 @@ func execCreateSalsaProject(args []string) {
 		log.Fatalf("Could not create webhook on salsa project: %s\n", err)
 	}
 
+	for _, branch := range []string{"master", "debian/*", "upstream", "pristine-tar"} {
```
stapelberg: Could you add upstream/* in preparation for adopting DEP14 please?

aviau: done!

aviau:
I have added upstream/*  to the protected branches.

PR #86, opened 2018-02-04
Update make.go - Changing the standards version

rajudev:
Updating the standards version to follow the latest as per the latest standards version of Debian policy. 
`4.1.3`


rajudev:
the PR fixes it.

PR #90, opened 2018-02-27
Remove shlibs:Depends from library packages

elboulangero:
Currently, `dh-make-golang` automatically adds `${shlibs:Depends}` in the `Depends` field of the library package. However when building the package, it leads to this warning:

    unknown substitution variable ${shlibs:Depends}

So it looks like field `${shlibs:Depends}` shouldn't be there.

### DIFF
--- a/make.go
+++ b/make.go
@@ -502,10 +502,11 @@ func writeTemplates(dir, gopkg, debsrc, debbin, debversion, pkgType string, depe
 	fmt.Fprintf(f, "Testsuite: autopkgtest-pkg-go\n")
 	fmt.Fprintf(f, "\n")
 	fmt.Fprintf(f, "Package: %s\n", debbin)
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for the PR!

I looked into this some more to understand precisely when shlibs:Depends is used.

IIUC, `dh_shlibdeps` locates all dynamically linked binaries and calls `dpkg-shlibdeps` on them, which in turn writes the substvar.

In Go, binaries are statically linked by default, unless you use the `os/user` package, DNS resolving, or any explicit cgo dependencies (e.g. sqlite3). Hence, not even all binary packages need shlibs:Depends, but figuring out whether a package needs it is much more work than just leaving it there and ignoring the warning.

Anyway, library packages definitely don’t need it, so I’m going ahead and merging your PR!

PR #95, opened 2018-04-18
Allow passing a commit to -git_revision

elboulangero:
Right now `dh-make-golang make` fails when we pass a commit to `-git_revision`. For example:

```
$ dh-make-golang make -git_revision aaff9d5  github.com/moby/buildkit
2018/04/18 09:24:36 Downloading "github.com/moby/buildkit/..."
2018/04/18 09:24:41 Deleting upstream vendor/ directories
2018/04/18 09:24:41 Determining upstream version number
fatal: No tags can describe 'aaff9d591ef128560018433fe61beb802e149de8'.
Try --always, or create some tags.
2018/04/18 09:24:41 Could not create a tarball of the upstream source: exit status 128
```

This is due to the command `git describe --long --all` (in `version.go`) which fails when git is in detached state.

This commit fixes that, and uses `git rev-parse --short HEAD` instead, as a last attempt to get a description of the current commit.

### DIFF
--- a/version.go
+++ b/version.go
@@ -51,26 +51,28 @@ func pkgVersionFromGit(gitdir string) (string, error) {
 	// This results in an output like 4.10.2-232-g9f107c8
 	cmd = exec.Command("git", "describe", "--long", "--tags")
 	cmd.Dir = gitdir
+	lastCommitSha := ""
 	describeBytes, err := cmd.Output()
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for the pull request, and sorry for the long delay! Just now had a chance to test this.

PR #102, opened 2018-07-27
use  team+pkg-go@tracker.debian.org in Maintainer

aviau:
According to
 - https://wiki.debian.org/Salsa/AliothMigration#tracker.debian.org
 - https://qa.pages.debian.net/distro-tracker/usage/teams.html#team-email-address

this is the email that we should use for the team at https://tracker.debian.org/teams/pkg-go/.

> This email address can also be used in the Maintainer field of Debian packages. By doing this, you ensure that your team’s packages are automatically added to the Distro Tracker team and that all members get the associated messages.

Users can modify emails that they want to receive for the team at https://tracker.debian.org/accounts/subscriptions/ with the Modify Keyword button.

### DIFF
--- a/make.go
+++ b/make.go
@@ -491,7 +491,7 @@ func writeTemplates(dir, gopkg, debsrc, debbin, debversion, pkgType string, depe
 	// TODO: change this once we have a “golang” section.
 	fmt.Fprintf(f, "Section: devel\n")
 	fmt.Fprintf(f, "Priority: optional\n")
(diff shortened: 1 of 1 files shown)


aviau:
The subject was discussed several times in debian-devel.

As far as migration goes, there is no need to modify exisitng packages as tracker.debian.org supports tracking packages that use the alioth mailing list too. Existing teams and packages were imported in the distro tracker.

I have cc-ed the pkg-go lists in a mail where I mention this (Changing the team's maintainer address).

PR #103, opened 2018-08-05
Arno/misc

elboulangero:
Details


elboulangero:
Actually I'm not so sure about adding `log.Printf("    git push\n")`.

Can someone tell me if following the `dh-make-golang make` instructions work out of the box? For me, the `master` branch was not pushed on salsa, only the `upstream` and `pristine-tar`.

stapelberg:
> Actually I'm not so sure about adding log.Printf(" git push\n").
> 
> Can someone tell me if following the dh-make-golang make instructions work out of the box? For me, the master branch was not pushed on salsa, only the upstream and pristine-tar.

`gbp push` pushes all branches, there is no need for a separate `git push`. Can you remove that change from the pull request, so that I can merge the others? If you observe any issues with `gbp push`, we should investigate that, but in a separate issue :).

Thanks!

elboulangero:
Hi Michael,

I removed the change from the PR, you can merge. Next time I happen to create a salsa repo I will be extra careful and see if there's a problem with `gbp push` or not.

Thanks!

PR #106, opened 2018-08-27
Check if name or email are present on global level before adding to git locally

lechner:
Untested, but is it approximately what you had in mind? Thank you!

### DIFF
--- a/make.go
+++ b/make.go
@@ -279,6 +279,26 @@ func runGitCommandIn(dir string, arg ...string) error {
 	return cmd.Run()
 }
 
+func execGitConfig(args ...string) (string, error) {
+	gitArgs := append([]string{"config", "--get", "--null"}, args...)
+	var stdout bytes.Buffer
+	cmd := exec.Command("git", gitArgs...)
+	cmd.Stdout = &stdout
+	cmd.Stderr = ioutil.Discard
+
+	err := cmd.Run()
+	if exitError, ok := err.(*exec.ExitError); ok {
+		if waitStatus, ok := exitError.Sys().(syscall.WaitStatus); ok {
+			if waitStatus.ExitStatus() == 1 {
+				return "", &ErrNotFound{Key: args[len(args)-1]}
+			}
+		}
+		return "", err
+	}
+
+	return strings.TrimRight(stdout.String(), "\000"), nil
+}
+
 func createGitRepository(debsrc, gopkg, orig string) (string, error) {
 	wd, err := os.Getwd()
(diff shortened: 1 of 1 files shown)


make.go
```diff
@@ -279,6 +279,26 @@ func runGitCommandIn(dir string, arg ...string) error {
 	return cmd.Run()
 }
 
+func execGitConfig(args ...string) (string, error) {
+	gitArgs := append([]string{"config", "--get", "--null"}, args...)
+	var stdout bytes.Buffer
+	cmd := exec.Command("git", gitArgs...)
+	cmd.Stdout = &stdout
+	cmd.Stderr = ioutil.Discard
```
stapelberg: This seems redundant — not setting cmd.Stderr has the same effect, no?

make.go
```diff
@@ -279,6 +279,26 @@ func runGitCommandIn(dir string, arg ...string) error {
 	return cmd.Run()
 }
 
+func execGitConfig(args ...string) (string, error) {
+	gitArgs := append([]string{"config", "--get", "--null"}, args...)
+	var stdout bytes.Buffer
+	cmd := exec.Command("git", gitArgs...)
+	cmd.Stdout = &stdout
```
stapelberg: Instead of dealing with the buffer manually, use `cmd.Output()` instead of `cmd.Run()` below.

make.go
```diff
@@ -279,6 +279,26 @@ func runGitCommandIn(dir string, arg ...string) error {
...
+	cmd.Stderr = ioutil.Discard
+
+	err := cmd.Run()
+	if exitError, ok := err.(*exec.ExitError); ok {
+		if waitStatus, ok := exitError.Sys().(syscall.WaitStatus); ok {
+			if waitStatus.ExitStatus() == 1 {
+				return "", &ErrNotFound{Key: args[len(args)-1]}
```
stapelberg: We don’t have the ErrNotFound type, so just return err here

make.go
```diff
@@ -279,6 +279,26 @@ func runGitCommandIn(dir string, arg ...string) error {
 	return cmd.Run()
 }
 
+func execGitConfig(args ...string) (string, error) {
+	gitArgs := append([]string{"config", "--get", "--null"}, args...)
```
stapelberg: Can you add a comment as to why --null is required here?

lechner: Yes, I will. Per `git-config` manpage, it "allows for secure parsing of the output without getting confused e.g. by values that contain line breaks." Thank you!

PR #109, opened 2019-02-07
Packaging and template cleanups

guillemj:
This MR contains several packaging and template cleanups and improvements. Even though some of these can be considered stylistic changes, they serve specific practical purposes, as documented in their respective commits.


stapelberg:
Thanks!

PR #114, opened 2019-08-24
Use Go Modules

dawidd6:


anthonyfok:
Thank you @dawidd6, merged in commit 0f79a23

PR #115, opened 2019-08-24
Use go-isatty instead of C call

dawidd6:


anthonyfok:
Nice!  Thank you for your contribution!

PR #116, opened 2019-08-24
New proposed workflow changes

dawidd6:
- not using pristine-tar anymore
- `debian/sid` branch as default HEAD branch instead of `master`
- postclone hook in `gbp.conf`

https://go-team.pages.debian.net/workflow-changes.html

### DIFF
--- a/make.go
+++ b/make.go
@@ -293,6 +293,11 @@ func createGitRepository(debsrc, gopkg, orig string) (string, error) {
 		return dir, err
 	}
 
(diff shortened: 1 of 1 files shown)


anthonyfok:
Hi @dawidd6, thank you for spearheading the change to the new workflow!

Note that I did end up reverting the addition of

 [clone]
 postclone=origtargz

to debian/gbp.conf because the proposed workflow change is to add that to ~/.gbp.conf, not to debian/gbp.conf of every single Go package.  Thank you for your understanding!

PR #124, opened 2020-01-05
Refactor shortHostName

zhsj:
Move knownHosts to a map to simplify the code,
and fix the publicsuffix guess, input should be fqdn.

### DIFF
--- a/make.go
+++ b/make.go
@@ -537,48 +537,35 @@ func normalizeDebianProgramName(str string) string {
 }
 
 func shortHostName(gopkg string, allowUnknownHoster bool) (host string, err error) {
+
+	knownHosts := map[string]string{
+		"github.com":        "github",
+		"code.google.com":   "googlecode",
+		"cloud.google.com":  "googlecloud",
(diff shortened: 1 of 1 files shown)


make.go
```diff
@@ -537,48 +537,35 @@ func normalizeDebianProgramName(str string) string {
...
+		"golang.org":        "golang",
+		"google.golang.org": "google",
+		"gitlab.com":        "gitlab",
+		"bitbucket.org":     "bitbucket",
+		"bazil.org":         "bazil",
+		"blitiri.com.ar":    "blitiri",
+		"pault.ag":          "pault",
+		"howett.net":        "howett",
+		"go4.org":           "go4",
+		"salsa.debian.org":  "debian",
```
tianon: Would it make sense to also take the opportunity to alphabetize this list to help improve future maintainability?

zhsj: since it has been merged, I will update it in another PR.

zhsj: sort it in #125

PR #126, opened 2020-01-16
Add sourcehut (git.sr.ht) to knownHosts

ThreeFx:
In the interest of packaging [aerc](https://aerc-mail.org) ([ITP](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=947962#5)), I think it is prudent to add git.sr.ht as known host for Go packages.

I have no idea how `go get` works internally, maybe it would make more sense to add each sourcehut VCS with it's VCS name (git-sourcehut or sourcehut-git), since it will prevent conflicts in the future. Any input is greatly appreciated.


zhsj:
IMO it's fine to ignore other vcs services provided by sourcehut.. IIRC, the proportion of non-git Go packages is very very low.

PR #128, opened 2020-01-16
Normalize every part of a Gopkg's name

### LINKED ISSUE #127: Improve upstream-to-package-name translation
The [Debian package naming guidelines](https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_debian_package_file_names) specify that the only valid characters in package names are: `[a-zA-Z0-9.+-]`.

The VCS hosting service sourcehut (git.sr.ht) has a tilde in front of user's names, thus confusing `git-buildpackage` when importing the original source.

I have no idea how the "canonical" fix for this would look. I suggest that we simply call `normalizeDebianProgramName` on every part of `gopkg` in `debianNameFromGopkg` (as in #128).

ThreeFx:
Closes #127.

See my thoughts on the issue for deciding whether this is a good idea.

### DIFF
--- a/make.go
+++ b/make.go
@@ -596,7 +596,15 @@ func debianNameFromGopkg(gopkg string, t packageType, allowUnknownHoster bool) s
 	}
 	parts[0] = host
 
-	return strings.Trim("golang-"+strings.ToLower(strings.Replace(strings.Join(parts, "-"), "_", "-", -1)), "-")
+	for i := range parts {
+		if i == 0 {
+			continue
+		}
+
+		parts[i] = normalizeDebianProgramName(parts[i])
+	}
+
+	return strings.Trim("golang-"+strings.Join(parts, "-"), "-")
 }
 
 func getDebianName() string {


ThreeFx:
What is the status on this?

make.go
```diff
@@ -596,7 +596,15 @@ func debianNameFromGopkg(gopkg string, t packageType, allowUnknownHoster bool) s
 	}
 	parts[0] = host
 
-	return strings.Trim("golang-"+strings.ToLower(strings.Replace(strings.Join(parts, "-"), "_", "-", -1)), "-")
+	for i := range parts {
+		if i == 0 {
+			continue
+		}
+
+		parts[i] = strings.ToLower(normalizeDebianProgramName(parts[i]))
```
zhsj, line 604: `parts[i] = normalizeDebianProgramName(parts[i])`

No need `strings.ToLower`

ThreeFx: Fixed, thanks!

PR #132, opened 2020-02-11
minor code improvements, read description

kataras:
Fixes:

- ineffectual assignment to `err`
- `Pebi` and `Exbi` variables are never used
- Printf-style function with dynamic format string and no further arguments should use print-style.

@zhsj please review & accept.

Thanks,
Gerasimos Maropoulos

### DIFF
--- a/progress.go
+++ b/progress.go
@@ -15,8 +15,6 @@ const (
 	Mebi
 	Gibi
 	Tebi
-	// Pebi
-	// Exbi
 )
 
 func humanizeBytes(b int64) string {
--- a/progress.go
+++ b/progress.go
@@ -15,8 +15,8 @@ const (
 	Mebi
 	Gibi
 	Tebi
-	Pebi
-	Exbi
+	// Pebi
+	// Exbi
 )
 
 func humanizeBytes(b int64) string {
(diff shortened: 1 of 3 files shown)


progress.go
```diff
@@ -15,8 +15,8 @@ const (
 	Mebi
 	Gibi
 	Tebi
-	Pebi
-	Exbi
+	// Pebi
```
zhsj: I prefer just deleting them.

template.go
```diff
@@ -137,7 +137,7 @@ func writeDebianControl(dir, gopkg, debsrc, debLib, debProg string, pkgType pack
 	fmt.Fprintf(f, "Maintainer: Debian Go Packaging Team <team+pkg-go@tracker.debian.org>\n")
 	fprintfControlField(f, "Uploaders", []string{getDebianName() + " <" + getDebianEmail() + ">"})
 	// TODO: change this once we have a “golang” section.
-	fmt.Fprintf(f, "Section: devel\n")
+	fmt.Fprint(f, "Section: devel\n")
```
zhsj: why just modify this line, there are other lines call `fmt.Fprintf` without format. It's better to keep the code consistent.

progress.go
kataras: I totally agree.

template.go
kataras: Yes but there are only three that uses a `variable` to print, though I think it's better to modify those three to use `print-style` instead of `printf-style`. However, you are the boss, you choose.

zhsj:
Merged in ab7810ecffa1eae68db1c130a91667539b70b855

template.go
anthonyfok: @kataras When I first began touching dh-make-golang a few months ago, I used to agree with you too, and actually started to change some of the `fmt.Fprintf` to `fmt.Fprint`, but very soon I realized that @stapelberg the original author has been consistently using `fmt.Fprintf()`, most likely to keep the text aligned.  So I bet that's why both @zhsj came to the same conclusion and ended up not deviating from the convention set out by the original author.

Issue #89, opened 2018-02-04
Have an option to determine recursive package dependencies.


rajudev:
While running `dh-make-golang <package-name>` , we get a list of dependencies which are still not in Debian. 
Then we have to do the same thing for all the dependencies and there sub-dependencies on multiple hierarchies manually. 

It would be great if dh-make-golang has an optional flag, may be `-r` or something like `--recursive-dependencies` so that it can generate the whole dependency tree. 

A simple text hierarchical output of the dependency tree is fine as well. The functionality can be further enhanced to provide an image file of the dependency tree, some thing like [this](https://people.debian.org/~boutil/diaspora/diaspora_deps20150721.pdf)

stapelberg:
Implemented the estimate command, which should do the trick. Leaving this issue open until -git_revision is supported. Issue #88 is a somewhat related usability improvement.

rajudev:
Like, already. I was planning to start working on this. 

Anyways, it would have taken me ages. Glad you implemented it. 
Will test the functionality in sometime.

stapelberg:
`dh-make-golang estimate github.com/jinzhu/gorm` doesn’t list the test-only dependency github.com/erikstmartin/go-testdb (packaged in golang-github-erikstmartin-go-testdb-dev).

I think we’ll need to add TestImports and XTestImports to the `go list` template.

rajudev:
Is the `go estimate <package-name>` reliable for use now?
Also while estimating dependencies for a package. 
I just got a list of packaged and un-packaged dependencies, instead of the tree graph that we planned. 
A list just gets which packages are not yet done, without letting me know which package depends on what. 
Am I missing something, or that is yet to be implemented.

stapelberg:
The estimate command is ready to use.

Which package are you running this on that a tree is required instead of a list?

rajudev:
Well running it for http://github.com/zyedidia/micro
It has `dependencies` and `dependencies has there  own dependencies` . 
I have already determined the dependency tree manually. It is just that I am testing the `estimate` option if it could have saved me the manual work.

rajudev:
Using the estimate command complains for the `digraph` tool not being present. 
May be we should add `golang-golang-x-tools` as a dependency for `dh-make-golang` in control. 
```
rajudev@5658cb08429d:~/dev/salsa.debian.org$ dh-make-golang estimate github.com/Debian/dh-make-golang
/bin/sh: 1: digraph: not found
[...]

PR #137, opened 2020-03-08
Add support for recursive dep graph

ramantehlan:
This PR is to address Issue #89 and to add support to check recursively if the project and test dependencies are packaged for Debian or not. 

- It will add a `-r` sub-flag to current `estimate` flag, which will be separate from plain estimate command.
- It will add a `-format` sub-flag, to supports three output formats **("Graph", "Map", "List")**

### Usage
```
$ dh-make-golang estimate -r github.com/zyedidia/micro/cmd/micro

$ dh-make-golang estimate -format map -r github.com/zyedidia/micro/cmd/micro
```

I have also created a standalone tool for it, which you can check out [here](https://github.com/ramantehlan/GoCheckDeb).

### DIFF
--- a/estimate.go
+++ b/estimate.go
@@ -175,27 +175,25 @@ func estimate(importpath string) error {
 	return nil
 }
 
-func recursiveEstimate(importpath string) error {
+func recursiveEstimate(importpath string, formatType string) error {
 	log.Print(aurora.Bold(aurora.Green("[DebGoGraph Starting]")))
 
-	project := importpath
-	returnType := "graph"
 	debFilter := true
 	displayAll := true
 
-	fmt.Printf("\nGolang package: %s\n", aurora.Bold(project))
-	fmt.Printf("Output type: %s\n", aurora.Bold(returnType))
+	fmt.Printf("\nGolang package: %s\n", aurora.Bold(importpath))
+	fmt.Printf("Output type: %s\n", aurora.Bold(formatType))
 	fmt.Printf("Deb filter: %v\n", aurora.Bold(debFilter))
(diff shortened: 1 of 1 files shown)


ramantehlan:
@rajudev can you review, or assign it to the maintainer for the review. For some reason Github don't show me that option anymore.

utkarsh2102:
Hi @anthonyfok,
Could you take a look at this?

estimate.go
```diff
@@ -173,8 +175,44 @@ func estimate(importpath string) error {
...
+
+	fmt.Printf("\nGolang package: %s\n", aurora.Bold(importpath))
+	fmt.Printf("Output type: %s\n", aurora.Bold(formatType))
+	fmt.Printf("Deb filter: %v\n", aurora.Bold(debFilter))
+	fmt.Printf("Display all deb (false for only main package): %v\n\n", aurora.Bold(displayAll))
+
+	fmt.Printf("Fetching dependencies of %s | It may take a while.\n\n", aurora.Bold(aurora.BrightBlue(importpath)))
+
+	// List | Graph | Tree
+	m, err := gocheckdeb.GetDep(importpath, "imports", formatType)
```
zhsj: I'm not a fan of using many third-party libraries. Especially this library has some overlap functions with dh-make-golang.

estimate.go
```diff
@@ -12,6 +12,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/logrusorgru/aurora"
```
zhsj: this library is not packaged in debian too. it's not so important to have colorful output...

n-peugnet:
This can be closed, now that #240 is merged.

PR #146, opened 2020-11-29
estimate: Fix recursion and improve var naming

imsodin:
The fix/functional change is in lines 135 and following: The recursion did nothing before, as `visit` is the default (noop) function at that point. You need a separate function variable to get actual recursion.

The rest is just variable renaming (took too many brain cycles to understand what's going on there ;) ) and filtering unrelated forward dependencies in the first loop.

### DIFF
--- a/estimate.go
+++ b/estimate.go
@@ -56,7 +56,7 @@ func removeVendor(gopath string) (found bool, _ error) {
 	return found, err
 }
 
-func estimate(importpath string) error {
+func estimate(baseImportPath string) error {
 	// construct a separate GOPATH in a temporary directory
 	gopath, err := ioutil.TempDir("", "dh-make-golang")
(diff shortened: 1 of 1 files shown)


zhsj:
> The fix/functional change is in lines 135 and following: The recursion did nothing before, as `visit` is the default (noop) function at that point. You need a separate function variable to get actual recursion.

I don't understand this. I write a poc code, it seems to work. https://play.golang.org/p/P5GUXzQ1jOi

imsodin:
You are clearly right, I don't understand what I was thinking back then. I remember having an example, but it clearly must have been wrong. The variable in the closure is evaluated when called, so it isn't a problem at all. Sorry for wasting your time (if you would like to keep the naming changes, let me know and I'll remove the bogus stuff).

Issue #107, opened 2018-12-02
Output directive will cause invalid initial push to salsa-project


suntong:
(Converted to issue from https://lists.debian.org/debian-go/2018/12/msg00001.html)

Summary: the output of `dh-make-golang` that gives instructions on what do next will cause invalid initial push to salsa-project.

Here is what I'm trying to follow (The whole log is at http://paste.debian.net/1053985/)

```
$ dh-make-golang github.com/danverbraganza/varcaser
2018/11/25 09:07:42 To create the packaging git repository on salsa, use:
2018/11/25 09:07:42     dh-make-golang create-salsa-project
golang-github-danverbraganza-varcaser
2018/11/25 09:07:42
2018/11/25 09:07:42 Once you are happy with your packaging, push it to
salsa using:
2018/11/25 09:07:42     git remote set-url origin
git@salsa.debian.org:go-team/packages/golang-github-danverbraganza-varcaser.git
2018/11/25 09:07:42     gbp push
                        ^^^^^^^^
```

Here is my work log:

```
$ git branch -v
* master       9078e4c - [+] Initial packaging
  pristine-tar fbba72a pristine-tar data for
golang-github-danverbraganza-varcaser_0.0~git20151108.ce61ec4.orig.tar.xz
  upstream     88a5d5c New upstream version 0.0~git20151108.ce61ec4

$ gbp push --dry-run --verbose
gbp:debug: ['git', 'rev-parse', '--show-cdup']
gbp:debug: ['git', 'rev-parse', '--is-bare-repository']
gbp:debug: ['git', 'rev-parse', '--git-dir']
gbp:debug: ['git', 'symbolic-ref', 'HEAD']
gbp:debug: ['git', 'show-ref', 'refs/heads/master']
gbp:debug: ['git', 'config', 'branch.master.remote']
gbp:debug: ['git', 'tag', '-l', 'debian/0.0_git20151108.ce61ec4-1']
gbp:debug: ['git', 'tag', '-l', 'upstream/0.0_git20151108.ce61ec4']
gbp:debug: ['git', 'rev-parse', '--quiet', '--verify',
'upstream/0.0_git20151108.ce61ec4^{commit}']
gbp:debug: ['git', 'rev-parse', '--quiet', '--verify', 'refs/heads/upstream']
gbp:debug: ['git', 'show-ref', '--verify', 'refs/heads/pristine-tar']
gbp:debug: ['git', 'log', '--pretty=format:%H', '--no-merges',
[...]

PR #153, opened 2021-01-19
Fix 107

elboulangero:
Refer to #107 for the reason why. I believe this fix will is helpful for newcomers.

### DIFF
--- a/make.go
+++ b/make.go
@@ -982,6 +982,7 @@ func execMake(args []string, usage func()) {
 	fmt.Printf("    dh-make-golang create-salsa-project %s\n", debsrc)
 	fmt.Printf("\n")
 	fmt.Printf("Once you are happy with your packaging, push it to salsa using:\n")
+	fmt.Printf("    git push origin %s\n", debBranch)
 	fmt.Printf("    gbp push\n")
 	fmt.Printf("\n")
 
--- a/make.go
+++ b/make.go
@@ -382,7 +382,7 @@ func runGitCommandIn(dir string, arg ...string) error {
 }
 
 func createGitRepository(debsrc, gopkg, orig string, u *upstream,
-	includeUpstreamHistory, allowUnknownHoster, dep14, pristineTar bool) (string, error) {
+	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string, pristineTar bool) (string, error) {
 	wd, err := os.Getwd()
 	if err != nil {
 		return "", err
@@ -392,14 +392,8 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 		return "", err
 	}
 
-	if err := runGitCommandIn(dir, "init"); err != nil {
+	if err := runGitCommandIn(dir, "init", "-b", debianBranch); err != nil {
 		return dir, err
-	}
-
-	if dep14 {
-		if err := runGitCommandIn(dir, "checkout", "-q", "-b", "debian/sid"); err != nil {
-			return dir, err
-		}
 	}
 
 	// Set repository options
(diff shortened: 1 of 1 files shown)


elboulangero:
Anyone willing to review this one?

creekorful (approved):

make.go
```diff
@@ -392,16 +392,10 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 		return "", err
 	}
 
-	if err := runGitCommandIn(dir, "init"); err != nil {
+	if err := runGitCommandIn(dir, "init", "-b", debianBranch); err != nil {
```
creekorful, line 395: This option has been added in Git v2.28.0, I'm wondering if it wouldn't be better to instead run `git checkout` later on in order to support older Git version. What do you think?

If the end users only install dh-make-golang through the archive this change won't matter (unstable is: 2.32.0), but I'm a bit concerned.

creekorful (changes requested):

make.go
elboulangero: > This option has been added in Git 2.28.0

Good catch, I completely missed that! Thanks for checking.

> I'm wondering if it wouldn't be better to instead run git checkout later on in order to support older Git version

So you mean running `git init && git checkout -b <debian-branch>` right? Seems like it's equivalent indeed, let me update the patch.

elboulangero: Done. I commented the code with "init -b", so that in the future if we want to use it, it's just a matter of un-commenting it.

creekorful (approved):

make.go
```diff
@@ -991,6 +991,7 @@ func execMake(args []string, usage func()) {
 	fmt.Printf("    dh-make-golang create-salsa-project %s\n", debsrc)
 	fmt.Printf("\n")
 	fmt.Printf("Once you are happy with your packaging, push it to salsa using:\n")
+	fmt.Printf("    git push origin %s\n", debBranch)
 	fmt.Printf("    gbp push\n")
```
creekorful, line 995: Looks good to me, however I'm still not sure about `gbp push`. As per the [man page](https://manpages.debian.org/buster/git-buildpackage/gbp-push.1.en.html) and the [bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=908204#10), it is clearly indicated that `gbp push` is intended to be run *after uploading a Debian package to the archive*. In the bug report Guido suggest simply running `git push` when there was no release.

Adding a `--tips` flag to gbp push has been suggested [here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=908204#42) but looks not implemented atm.

Personally, I'm doing a `git push --all && git push  --tags` in order to push all the branches and the tags to the Salsa repository. I wonder what the rest of the team think about that. @zhsj any opinion on this?

zhsj: It's clear that the `gbp push` workflow doesn't work here. However I would much prefer implementing a `dh-make-golang push` command. And I don't like `git push --tags` since it pushes all upstream tags which are not needed, for example upstream have v1.0.0 -> v10.0.0 tags, while we only need upstream/10.0.0 tag.

creekorful: Ack, so this `dh-make-golang push` command should be running a `git push origin` to push the branches and then push all the tags beginning by `upstream/` ?

zhsj: Probably like `gbp push`, but allow to push wip debian branch: 

+ `git push origin debian/sid`
+ `git push origin upstream`
+ `git push origin upstream/<tag>`
+ (optional) `git push origin pristine-tar`
+ (optional) `git push origin debian/<tag>`

While I'm writing above, I find it forces `dh-make-golang` to parse `gbp.conf` to see the `upstream` branch name ..

PR #154, opened 2021-01-24
Packaging cleanups

guillemj:
These are some of the things I've been doing lately after generating the initial packaging.

### DIFF
--- a/make.go
+++ b/make.go
@@ -495,7 +495,7 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 		}
 		// Beginning newline in case the file already exists and lacks a newline
 		// (not all editors enforce a newline at the end of the file):
-		if _, err := f.Write([]byte("\n.pc\n")); err != nil {
+		if _, err := f.Write([]byte("\n/.pc/\n/_build/\n")); err != nil {
 			return dir, err
 		}
 		if err := f.Close(); err != nil {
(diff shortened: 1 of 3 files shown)


creekorful (approved):

make.go
```diff
@@ -495,7 +495,7 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 		}
 		// Beginning newline in case the file already exists and lacks a newline
 		// (not all editors enforce a newline at the end of the file):
-		if _, err := f.Write([]byte("\n.pc\n")); err != nil {
+		if _, err := f.Write([]byte("\n/.pc/\n/_build/\n")); err != nil {
```
creekorful, line 498: `"\n/.pc/\n/_build/\n"` this looks rather strange no?
shouldn't it be `"\n.pc\n_build\n"` ?

creekorful (changes requested):

make.go
guillemj: This tells git to ignore these if they are directories and are sitting at the root of the repository. Otherwise this will show them as untracked, as those would be unusual files to have, and conflict with the expectations of the tools.

creekorful: Didn't know that! Looks good to me then.

creekorful (approved):

creekorful:
Thanks!

PR #156, opened 2021-02-10
Bump to latest Debian Policy version

fauust:
See: https://www.debian.org/doc/debian-policy/


creekorful (approved):

guillemj:
Already covered as part of #154 . :)

fauust:
OK indeed! Closing then :)

PR #171, opened 2021-10-19
Implement check-depends

creekorful:
Hi, here's the check-depends feature! (finally)

I have several points I wanna have a review on:

- I use the go.mod to parse Go dependencies, this mean I won't work for non Go module package. Is there a lot of such package on the archive?


creekorful:
Hello @anthonyfok, @zhsj do you mind reviewing the PR? 😄 

I have several points I wanna have an advice on:

- I use the go.mod to parse Go dependencies, this mean I won't work for non Go module package. Is there a lot of such package on the archive?
- When running `check-depends` on dh-make-golang I have the following output:

```
NEW dependency github.com/google/go-github/v38 is NOT yet packaged in Debian
RM dependency  (golang-github-google-go-github-dev)
```

But indeed the library is packaged, but under `github.com/google/go-github` import path. How should I approach that? Should I consider that finding a package with import path that match has prefix the current one good?

zhsj:
> Hello @anthonyfok, @zhsj do you mind reviewing the PR? smile
> 
> I have several points I wanna have an advice on:
> 
> * I use the go.mod to parse Go dependencies, this mean I won't work for non Go module package. Is there a lot of such package on the archive?
> * When running `check-depends` on dh-make-golang I have the following output:
> 
> ```
> NEW dependency github.com/google/go-github/v38 is NOT yet packaged in Debian
> RM dependency  (golang-github-google-go-github-dev)
> ```
> 
> But indeed the library is packaged, but under `github.com/google/go-github` import path. How should I approach that? Should I consider that finding a package with import path that match has prefix the current one good?

I think it's OK to support Go module only. But by parsing go.mod file means we have two algorithms in dh-make-golang to determine the Go dependencies.

The origin one:

https://github.com/Debian/dh-make-golang/blob/cd2e4303445a360af9c8c531bbde05624c2af5db/make.go#L257

Below this function,

https://github.com/Debian/dh-make-golang/blob/cd2e4303445a360af9c8c531bbde05624c2af5db/make.go#L321

It tries to use `vcs.RepoRootForImportPath` to determine the repository path, then 

https://github.com/Debian/dh-make-golang/blob/cd2e4303445a360af9c8c531bbde05624c2af5db/make.go#L982

It converts the package name for the dependencies.

I assume the origin code is able to handle `github.com/google/go-github/v38`.

creekorful:
Hello @zhsj,

I have used `vcs.RepoRootForImportPath` and now the function is able to correctly resolve repository import path.
As for the two algorithms used to parse dependencies I think it's not the worst scenario to have these two algorithms because:

- The first one is able to work with non Go module (so we are still able to package them)
- The other one may only work for Go module but there's a **LOT** less boilerplate code, it will be compatible with never Go version (and the first one won't).

I think we targeting newer version of Go the first algorithm may be replaced with the second one.

What do you think?

creekorful:
We will go like this for v1 then and will improve with usages :)

PR #175, opened 2021-11-29
Use new dh-sequence-golang instead of dh-golang and --with=golang

guillemj:
This simplifies the packaging, by only declaring this usage once.


anthonyfok:
This is great!  Thank you @guillemj!

PR #177, opened 2021-12-02
Generate a debian/.gitignore file

guillemj:
This helps ignoring artifacts that git should not track, and which it can safely clean.


anthonyfok:
Nice!  Thank you @guillemj!

PR #178, opened 2021-12-04
Reorder generated fields

guillemj:
While stylistic, the rationale for these changes is as follows:

= debian/control

- Group Section and Priority together as these categorize the package, and place them after Source or Package fields, as this determines where in the archive these might end up.
- Place Maintainer/Uploader after these, as the responsible party for the package.
- Move Rules-Requires-Root before Build-Depends, as this determines how to drive the building process, and might impose additional dependencies, such as fakeroot or sudo.
- Move the Testsuite after the Build-Depends, as this is part of the runnable metadata.

= debian/copyright

- Move the Source field after Format, as both line up, refer to an URL, and it's the first reference to the upstream project.

These tend to be the changes I do on packages I've ITPed, as they jump on my face pretty quickly. :smile:

### DIFF
--- a/template.go
+++ b/template.go
@@ -190,11 +190,11 @@ func writeDebianControl(dir, gopkg, debsrc, debLib, debProg string, pkgType pack
 	// Source package:
 
 	fmt.Fprintf(f, "Source: %s\n", debsrc)
+	fmt.Fprintf(f, "Section: golang\n")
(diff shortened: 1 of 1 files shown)


creekorful:
Thanks for the changes and the well detailed rationale :D

PR #189, opened 2022-12-06
make: download tarballs from sr.ht

supertassu:
and add a test for the tarball URL generation.


anthonyfok:
Thank you @supertassu!

PR #224, opened 2024-08-29
Update fields emitted

guillemj:
This series updates the `Standards-Version` field version, and switches from the `Built-Using` to the `Static-Built-Using` field, following the migration done in the `dh-golang` package.


creekorful:
Thank you!

Issue #226, opened 2024-11-27
"-git_revision" option of "make" subcommand results in unexpected package version when the upstream repo has two tags at the same commit


wlyu2:
When I try to create a Debian package for "k8s.io/cli-runtime", I get the following output:
```
# dh-make-golang make -git_revision v0.31.3 -type l k8s.io/cli-runtime                                                                                                    
2024/11/26 14:33:52 Starting "dh-make-golang v0.7.0 linux/amd64"          
2024/11/26 14:33:52 Downloading "k8s.io/cli-runtime/..."                                                                                                                                                    
2024/11/26 14:33:54 Determining upstream version number             
2024/11/26 14:33:54 Found latest tag "kubernetes-1.31.3"                   
2024/11/26 14:33:54 WARNING: Latest tag "kubernetes-1.31.3" is not a valid SemVer version
2024/11/26 14:33:54 Latest tag "kubernetes-1.31.3" matches master  
2024/11/26 14:33:54 Package version is "1.31.3"
```
When I check the upstream repo "k8s.io/cli-runtime", I observe that both tags `kubernetes-1.31.3" and "v0.31.3" refers to the same commit:
```
$ git branch
* (HEAD detached at v0.31.3)
  master
  
$ git log -1
commit 597e981c50a2b9cc2f484e6fc94f618822b62888 (HEAD, tag: v0.31.3, tag: kubernetes-1.31.3)
Author: Kubernetes Publisher <k8s-publishing-bot@users.noreply.github.com>
Date:   Thu Nov 21 05:10:54 2024 +0000

    Update dependencies to v0.31.3 tag
`
$ git remote -v
origin  https://github.com/kubernetes/cli-runtime.git (fetch)
origin  https://github.com/kubernetes/cli-runtime.git (push)
```

I assume the behavior of "dh-make-golang" is because 
- tag `kubernetes-1.31.3` is created at a later time after `v0.31.3` thus making it the latest tag before the `HEAD`.
-  "dh-make-golang" uses "git describe" which fetches the latest tag before the `HEAD` [1] which is checked out to the value specified by option "-git_revision" by "dh-make-golang" [2].

I would hope that, if the git revision specified by "-git_revision" is a valid tag, "dh-make-golang" should be using that tag without any additional heuristics because:
-  This, I think, would be more intuitive.
[...]

PR #227, opened 2024-11-27
dh-make-golang make: directly use the user preferred tag specified in "-git_revision" option

wlyu2:
When a user specifies a valid tag when using

    dh-make-golang make -git_revision {some valid tag}

if there is another tag pointing to the same commit as the specified tag, then dh-make-golang might use the other tag to determine the package version.

This patch modifies the behavior of "dh-make-golang make" such that, if the user specifies a valid tag in "-git_revision" option, "dh-make-golang make" will always use the specified tag to determine the upstream package version.

This patch also trivially changes the tests written in version_test.go so that the tests adjusts to the new function signature of version.go:pkgVersionFromGit().

### DIFF
--- a/version.go
+++ b/version.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


wlyu2:
This patch attempts to fix #226 .

version.go
```diff
@@ -32,16 +33,32 @@ var (
...
+	var cmd *exec.Cmd
+
+	// If the user specifies a valid tag as the preferred revision, that tag should be used without additional heuristics.
+	if u.rr != nil {
+		if out, err := u.rr.VCS.Tags(gitdir); err == nil && slices.Contains(out, preferredRev) {
+			latestTag = preferredRev
+			goto FoundLatestTag
+		}
+	}
+
```
wlyu2, line 51: These are some comments regarding why I use a `goto` statement. Please let me know if there are better alternatives.

I want to use the `git describe` heuristics as a fallback in case I cannot verify if the user specified git revision is valid or not.

The logic I want to implement is:
```
if (u.rr is valid (which makes u.rr.VCS.Tags meaningful) and the user specified tag exists as a valid tag) {
    set the latest tag to this user specified tag
} else {
    the fallback heuristics
}
```

However, to implement the "if" part of the above pseudocode, I cannot find a way to implement it as a single if clause. This will result in the "else" part being written out twice, and I want to avoid such duplication.

In other words, I am trying to deal with the problem in this post: https://stackoverflow.com/questions/13197425/multiple-if-with-one-else

William

ottok:
Does @n-peugnet or @Maytha8 want to review/approve this?

Maytha8 (approved):

wlyu2:
Closes #226

version.go
NightTsarina: Wouldn't it be better to set a `foundTag` flag or reusing the `if len(latestTag) > 0` test?

NightTsarina (approved):

Issue #228, opened 2024-11-28
gbp reporting error "Can't find upstream vcs tag" when the upstream tag is not on any branch


wlyu2:
When I am trying to create a debian package for "k8s.io/cli-runtime", I encountered the following error:
```
$ dh-make-golang make -git_revision v0.30.1 -type l k8s.io/cli-runtime
2024/11/26 10:33:32 Starting "dh-make-golang v0.7.0 linux/amd64"                                                                                                                                            
2024/11/26 10:33:32 revision: "v0.30.1"                                                                                                                                                                     
2024/11/26 10:33:32 Downloading "k8s.io/cli-runtime/..."                                                                                                                                                    
2024/11/26 10:33:34 Determining upstream version number                                                                                                                                                     
2024/11/26 10:33:34 Found latest tag "kubernetes-1.30.1"                                                                                                                                                    
2024/11/26 10:33:34 WARNING: Latest tag "kubernetes-1.30.1" is not a valid SemVer version                                                                                                                   
2024/11/26 10:33:34 Latest tag "kubernetes-1.30.1" matches master                                                                                                                                           
2024/11/26 10:33:34 Package version is "1.30.1"
...
2024/11/27 11:30:10 Adding remote "origin" with URL "git@salsa.debian.org:go-team/packages/golang-k8s-cli-runtime.git"                                                                                      
2024/11/27 11:30:10 Adding remote "k8s" with URL "https://github.com/kubernetes/cli-runtime"                                                                                                                
2024/11/27 11:30:10 Running "git fetch k8s"                                                           
remote: Enumerating objects: 8128, done.                                                              
remote: Counting objects: 100% (3177/3177), done.                                                                                                                                                           
remote: Compressing objects: 100% (643/643), done.                                                                                                                                                          
remote: Total 8128 (delta 2613), reused 3066 (delta 2525), pack-reused 4951 (from 1)                                                                                                                        
Receiving objects: 100% (8128/8128), 4.90 MiB | 17.98 MiB/s, done.                                    
Resolving deltas: 100% (5200/5200), done.                                                             
From https://github.com/kubernetes/cli-runtime                                                        
 * [new branch]      master                    -> k8s/master                                                                                                                                                
 * [new branch]      release-1.12              -> k8s/release-1.12                                                                                                                                          
 * [new branch]      release-1.13              -> k8s/release-1.13                                                                                                                                          
 * [new branch]      release-1.14              -> k8s/release-1.14                                                                                                                                          
[...]

PR #229, opened 2024-11-28
fix release tags not recognized when not on a branch

wlyu2:
"dh-make-golang" creates the "debianized" source repo from scratch. It creates an empty git repo, adds the upstream as a remote, and fetches the tags. However, before this patch, it is using "git fetch" with no additional flags or options. This results in upstream tags not on any branches not being picked up. This patch adds "--tags" flag to "git fetch" so that it can fetch all upstream tags.


wlyu2:
This merge request fixes #228 .

n-peugnet (approved):
I can confirm that it indeed fixes the issue. @wlyu2: you can add the content of your second comment in your original post, this will close the issue automatically on merge.

PR #225, opened 2024-11-20
Use DEP-14 branch names `debian/latest` and `upstream/latest`

ottok:
In DEP-14, the preferred branch name for the Debian packaging target branch is `debian/latest` and the preferred name for the upstream import target branch is `upstream/latest`. Note that the upstream development branch name can be whatever and should stay as it is upstream, typically `main` or `master`. The branch `upstream/latest` should not point to the latest upstream development commit, but to the latest commit that was used as the upstream release that the Debian revision was derived from.

PR #230, opened 2024-11-29
Stop modifying upstream .gitignore file and put build artifacts in `debian/build/`

ottok:
See commits for details. This was split out from #225.

You can see this is action at e.g.
* https://salsa.debian.org/go-team/packages/golang-github-muesli-mango-pflag
* https://salsa.debian.org/go-team/packages/golang-github-muesli-mango-cobra
* https://salsa.debian.org/go-team/packages/glow

### DIFF
--- a/template.go
+++ b/template.go
@@ -86,6 +86,7 @@ func writeDebianGitIgnore(dir, debLib, debProg string, pkgType packageType) erro
 	fmt.Fprintf(f, "*.log\n")
 	fmt.Fprintf(f, "*.substvars\n")
 	fmt.Fprintf(f, "/.debhelper/\n")
+	fmt.Fprintf(f, "/build/\n")
 	fmt.Fprintf(f, "/debhelper-build-stamp\n")
 	fmt.Fprintf(f, "/files\n")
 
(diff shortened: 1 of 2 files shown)


template.go
```diff
@@ -299,7 +300,7 @@ func writeDebianRules(dir string, pkgType packageType) error {
 	fmt.Fprintf(f, "#!/usr/bin/make -f\n")
 	fmt.Fprintf(f, "\n")
 	fmt.Fprintf(f, "%%:\n")
-	fmt.Fprintf(f, "\tdh $@ --builddirectory=_build --buildsystem=golang\n")
+	fmt.Fprintf(f, "\tdh $@ --builddirectory=debian/build --buildsystem=golang\n")
```
zhsj, line 303: NACK. There are already too many packages using that path.

ottok:
This is not an API. There isn't any downside for new packages to use this
clearly cleaner and better way, and old packages are trivial to update, and
testing that the update works correctly is also trivial. The different
gitignore/clean has been used before should not be an argument on its own.
Otherwise no progress could happen, right?

template.go
ottok: This is not an API. There isn't any downside for new packages to use this
clearly cleaner and better way, and old packages are trivial to update, and
testing that the update works correctly is also trivial. The different
gitignore/clean has been used before should not be an argument on its own.
Otherwise no progress could happen, right?

tianon: Maybe we should drop this flag entirely and update https://salsa.debian.org/go-team/packages/dh-golang/-/blob/81892fe8d3a51a22c910a649052669dd9f18837e/lib/Debian/Debhelper/Buildsystem/golang.pm#L275-279 instead?

ottok: Indeed, that would be ideal. Posted first version at https://salsa.debian.org/go-team/packages/dh-golang/-/merge_requests/25, perhaps we can continue the review of that change there?

If it is merged, this can be reduced to just drop those modifications as unnecessary.

Maytha8: @tianon Dropping the flag isn't a good idea, since any d/rules in existing packages rely on the build directory being where it is now when dh_auto_configure/dh_auto_build is overridden.

tianon: Sure, but this is the generator for new packages, not existing packages.  For existing packages where that matters, an explicit override is appropriate, and documented appropriately in `man dh-golang` (so this point matters more for some theoretical automated tool dropping the flag, which could be that it avoids dropping the flag in any case where it isn't set to this value, any case where `_build` appears elsewhere in the file too, and any case that fails to build after dropping the flag).

ottok: Isn't the optimal thing to merge both https://salsa.debian.org/go-team/packages/dh-golang/-/merge_requests/25 and this, so that the change is uniform everywhere and everyone gets the same result immediately? And once a new version dh-golang has been out there long enough, we can make dh-make-golang smaller and drop extra stuff from here.

ottok:
As the majority if in favor of this simplification, and it was also accepted and implemented in https://salsa.debian.org/go-team/packages/dh-golang/-/commit/bc16dff5381b668a71fa99c381baba202c34c789, I am now proceeding to merge this.

PR #239, opened 2025-01-05
Apply wrap-and-sort -at by default

n-peugnet:
The Go team currently recommends the flags --wrap-always --trailing-comma [1] which translates to -at, so dh-make-golang should use this flag combination by default.

[1] https://go-team.pages.debian.net/workflow-changes.html#wf-2017-11-debian-control

### DIFF
--- a/make.go
+++ b/make.go
@@ -827,7 +827,7 @@ func execMake(args []string, usage func()) {
 
 	fs.StringVar(&wrapAndSort,
 		"wrap-and-sort",
-		"a",
+		"at",
 		"Set how the various multi-line fields in debian/control are formatted.\n"+
 			"Valid values are \"a\", \"at\" and \"ast\", see wrap-and-sort(1) man page\n"+
 			"for more information.")


ottok:
This https://go-team.pages.debian.net/workflow-changes.html#wf-2017-11-debian-control is very old. I think most people nowadays would recommend `-ast`, which is also what [upstream wrap-and-sort maintainers seem](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=895570#66) to plan to make the future default for the whole tool, and what [debputy reformat](https://manpages.debian.org/unstable/dh-debputy/debputy.1.en.html) already uses.

So I am in favor of merging this but with `-ast` as the default.

ottok:
Based on recent discussions on mailing list and dh-make-golang and pkg-go-tools about workflows and defaults, and now also in https://go-team.pages.debian.net/workflow-changes.html#wf-2025-01-wrap-and-sort-ast, I think you should update this to default to `-ast`.

ottok:
Mailing list has also suggestion about `-astbk`:

> Simo: I believe for most
> cases it results in the same output, but it does result in predictable
> binary package sort order (except for the first package) which I find
> more predictable when paging through debian/control and having some
> mental model where to find different parts of a package.  Most packages
> in Go are simple with just one or two binary packages, so I'm hoping
> this is not too difficult to agree on if you already settled for '-tsa'.

NightTsarina (approved):

NightTsarina:
> So I am in favor of merging this but with `-ast` as the default.

I also prefer that combination, but it seems there is no consensus yet. I'd say let's merge this as is so we can release and start a thread about only this change.

PR #243, opened 2025-01-18
Add -git_revision flag to estimate command

n-peugnet:
Allows one to define the git revision of the specified Go package to estimate, defaulting to the default behavior of go get. Useful in case you do not want to estimate the latest version.

This was the reason why #89 was kept open (see https://github.com/Debian/dh-make-golang/issues/89#issuecomment-366081418), I overlooked that. So here it is.


ottok (approved):
LGTM and adding this feature is aligned with what original authors had planned.

NightTsarina (approved):

PR #244, opened 2025-01-21
Use current URL for CI badge

n-peugnet:
Just noticed this small discrepancy.


ottok:
Thanks, nice catch! I will merge this right away as it is just a typofix.

ottok (approved):

PR #245, opened 2025-01-25
Revert workflow changes that got rushed and merged with no Go team agreement nor consensus

guillemj:
This reverts several controversial changes.

This was pushed and merged in a rush without Go team agreement or consensus, based on false premises where even the DEP it is based on does not have any mention of there being defaults preferences for these changes, where there's been recorded objections in general in the Debian project at large in the past, and where even existing practice (as analyzed by the same author) does not support these defaults being the majority in use.

This should also unblock a pending upload to fix release critical bugs in the package in Debian.

Ref: https://lists.debian.org/debian-devel/2024/08/msg00299.html

### DIFF
--- a/make.go
+++ b/make.go
@@ -416,8 +416,7 @@ func runGitCommandIn(dir string, arg ...string) error {
 }
 
 func createGitRepository(debsrc, gopkg, orig string, u *upstream,
-	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string,
-	dep14 bool, pristineTar bool) (string, error) {
+	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string, dep14 bool, pristineTar bool) (string, error) {
 
 	// debianBranch is passed in function call, but upstream import branch needs
 	// also to be defined
--- a/make.go
+++ b/make.go
@@ -416,7 +416,7 @@ func runGitCommandIn(dir string, arg ...string) error {
 }
 
 func createGitRepository(debsrc, gopkg, orig string, u *upstream,
-	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string, dep14 bool, pristineTar bool) (string, error) {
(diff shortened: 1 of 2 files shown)


guillemj:
Cc: @zhsj, @creekorful, @elboulangero, @nileshpatra, @anthonyfok, @gibmat (I cannot add reviewers, and I'm probably missing people, feel free to add)

ottok:
> This was pushed and merged in a rush without Go team agreement or consensus

Can you please Guillem remove the sentence above? Based on reaction from multiple people seems that sentence is disparaging me, as people assume it is as a fact without second thoughts. I have deliberately been slow in changes and allowed discussions to take place on the mailing list, and in MRs/PRs, over several weeks and months. You know I replied this to you already 5 days ago in https://lists.debian.org/debian-go/2025/01/msg00123.html.

I would invite you to participate in the debian-go mailing list discussions. Here are a couple of pointers to key threads:
- Dec 5: https://lists.debian.org/debian-go/2024/12/msg00021.html
- Jan 2: https://lists.debian.org/debian-go/2025/01/msg00005.html
- Jan 13: https://lists.debian.org/debian-go/2025/01/msg00045.html

There are also other messages starting from mid-September, but I assume you didn't follow the list in the first place due to time limits, so those above should be enough to get context quickly, or just search emails with topic "Workflow changes proposal for 2025".

NightTsarina:
This LGTM, we can discuss this again after trixie is released, and please let's make sure we don't merge changes where it is clear there is no consensus again.

ottok:
Please update the PR description. If you have a false description and ask people to agree to false/misleading claims, then you are not actually getting agreement to the actual issue, but only support for perpetuating the false claims.

NightTsarina:
@ottok
The changes were indeed rushed and very explicit disagreements were ignored, the description is correct.
So please, stop. This is only hurting any chances of these changes getting accepted in the future and you are wasting valuable volunteer time.

PR #247, opened 2025-02-19
Use DEP-14 branch names debian/latest and upstream/latest, and upstream vcs remote name 'upstreamvcs'

ottok:
See commits for details. This is a re-submission of #225, pending to be merged potentially in the summer of 2025.

### DIFF
--- a/make.go
+++ b/make.go
@@ -419,6 +419,14 @@ func runGitCommandIn(dir string, arg ...string) error {
 func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string,
 	dep14 bool, pristineTar bool) (string, error) {
+
+	// debianBranch is passed in function call, but upstream import branch needs
+	// also to be defined
+	upstreamImportBranch := "upstream"
+	if dep14 {
+		upstreamImportBranch = "upstream/latest"
+	}
+
 	wd, err := os.Getwd()
 	if err != nil {
 		return "", fmt.Errorf("get cwd: %w", err)
@@ -464,7 +472,8 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 
 	// Preconfigure branches
 
-	branches := []string{debianBranch, "upstream/latest"}
+	branches := []string{debianBranch, upstreamImportBranch}
+
 	if pristineTar {
 		branches = append(branches, "pristine-tar")
 	}
@@ -478,13 +487,8 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 	}
 
 	if includeUpstreamHistory {
-		u.remote, err = shortHostName(gopkg, allowUnknownHoster)
-		if err != nil {
-			return dir, fmt.Errorf("unable to fetch upstream history: %q", err)
-		}
-		if u.remote == "debian" {
-			u.remote = "salsa"
-		}
+		// Always call the upstream git remote 'upstreamvcs' just like git-buildpackage does
+		u.remote = "upstreamvcs"
 		log.Printf("Adding remote %q with URL %q\n", u.remote, u.rr.Repo)
 		if err := runGitCommandIn(dir, "remote", "add", u.remote, u.rr.Repo); err != nil {
 			return dir, fmt.Errorf("git remote add %s %s: %w", u.remote, u.rr.Repo, err)
@@ -496,10 +500,13 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 	}
 
 	// Import upstream orig tarball
+	// (and release git tag if includeUpstreamHistory)
 
-	arg := []string{"import-orig", "--no-interactive", "--debian-branch=" + debianBranch}
-	if dep14 {
-		arg = append(arg, "--upstream-branch=upstream/latest")
+	arg := []string{
+		"import-orig",
+		"--no-interactive",
+		"--debian-branch=" + debianBranch,
+		"--upstream-branch=" + upstreamImportBranch,
 	}
 	if pristineTar {
 		arg = append(arg, "--pristine-tar")
--- a/make.go
+++ b/make.go
@@ -417,7 +417,8 @@ func runGitCommandIn(dir string, arg ...string) error {
 }
 
 func createGitRepository(debsrc, gopkg, orig string, u *upstream,
-	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string, pristineTar bool) (string, error) {
+	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string,
+	dep14 bool, pristineTar bool) (string, error) {
 	wd, err := os.Getwd()
 	if err != nil {
 		return "", fmt.Errorf("get cwd: %w", err)
@@ -463,7 +464,7 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 
 	// Preconfigure branches
 
-	branches := []string{debianBranch, "upstream"}
+	branches := []string{debianBranch, "upstream/latest"}
 	if pristineTar {
 		branches = append(branches, "pristine-tar")
 	}
@@ -497,6 +498,9 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 	// Import upstream orig tarball
 
 	arg := []string{"import-orig", "--no-interactive", "--debian-branch=" + debianBranch}
+	if dep14 {
+		arg = append(arg, "--upstream-branch=upstream/latest")
+	}
 	if pristineTar {
 		arg = append(arg, "--pristine-tar")
 	}
@@ -759,7 +763,7 @@ func execMake(args []string, usage func()) {
 	fs.BoolVar(&dep14,
 		"dep14",
 		true,
-		"Follow DEP-14 branch naming and use debian/sid (instead of master)\n"+
+		"Follow DEP-14 branch naming and use debian/latest (instead of master)\n"+
 			"as the default debian-branch.")
 
 	var pristineTar bool
@@ -869,7 +873,7 @@ func execMake(args []string, usage func()) {
 	// Set the debian branch.
 	debBranch := "master"
 	if dep14 {
-		debBranch = "debian/sid"
+		debBranch = "debian/latest"
 	}
 
 	switch strings.TrimSpace(wrapAndSort) {
@@ -960,7 +964,7 @@ func execMake(args []string, usage func()) {
 
 	debversion := u.version + "-1"
 
-	dir, err := createGitRepository(debsrc, gopkg, orig, u, includeUpstreamHistory, allowUnknownHoster, debBranch, pristineTar)
+	dir, err := createGitRepository(debsrc, gopkg, orig, u, includeUpstreamHistory, allowUnknownHoster, debBranch, dep14, pristineTar)
 	if err != nil {
 		log.Fatalf("Could not create git repository: %v\n", err)
 	}
--- a/template.go
+++ b/template.go
@@ -352,7 +352,8 @@ func writeDebianGbpConf(dir string, dep14, pristineTar bool) error {
 
 	fmt.Fprintf(f, "[DEFAULT]\n")
 	if dep14 {
-		fmt.Fprintf(f, "debian-branch = debian/sid\n")
+		fmt.Fprintf(f, "debian-branch = debian/latest\n")
+		fmt.Fprintf(f, "upstream-branch = upstream/latest\n")
 		fmt.Fprintf(f, "dist = DEP14\n")
 	}
 	if pristineTar {


NightTsarina (changes requested):
Most of the changes in this PR are not very controversial and I support them, even if I personally prefer `upstream/<upstream HEAD>` instead of `upstream/latest` (which is valid DEP-14 as we are *not* importing tarballs but following git history.

But big NACK to the change of the debian branch, you still haven't given a reason to change, except a misinterpretation of DEP-14. This PR cannot be merged until there is a consensus on this topic.

umlaeute:
> except a misinterpretation of DEP-14

how's `debian/latest` a *misinterpretation*?
i fully acknowledge that DEP-14 allows different names, and that there is controversy about which name to pick and that otto is pushing rather this controversial name strongly.
but nevertheless is `debian/latest` is explicitly suggested (among others) by DEP-14, so it's a valid interpretation.

guillemj:
What is a misinterpretation is stating the `debian/latest` is the preferred branch name by the DEP, which is not true.

ottok:
I did not invent `debian/latest` or `upstream/latest`, they are from DEP-14. I have however been using them successfully for many years and I think they make sense. Personally I would feel odd if I was using a branch called `debian/unstable` as the permanent default branch in git and the development branch, and occasionally uploading to experimental from a branch called 'unstable'. The reason DEP-14 (which I didn't write) recommends `debian/latest` is presumably to avoid mixing up default git branch name with the target archive name.

NightTsarina:
> I did not invent `debian/latest` or `upstream/latest`, they are from DEP-14. I have however been using them successfully for many years and I think they make sense. Personally I would feel odd if I was using a branch called `debian/unstable` as the permanent default branch in git and the development branch, and occasionally uploading to experimental from a branch called 'unstable'. The reason DEP-14 (which I didn't write) recommends `debian/latest` is presumably to avoid mixing up default git branch name with the target archive name.

@ottok Here lies the problem:

1. DEP-14 *does not* mandate `debian/latest`, `debian/<suite>` is also a valid option.
2. The team *does not* use `debian/unstable` for uploads to experimental, but `debian/experimental`.
3. Despite your experience, the go-team has been using DEP-14 successfully for many years, and there is absolutely no benefit in this change, but a huge cost.

rhansen:
As an outsider, this PR looks good to me.

I didn't see any specific objections to changing `upstream` to `upstream/<something>`, or to changing the upstream remote name to `uptreamvcs`.  Maybe just those changes (but with `upstream` renamed to `upstream/sid` for consistency with `debian/sid`?) would be less controversial?

>   1. DEP-14 _does not_ mandate `debian/latest`, `debian/<suite>` is also a valid option.

DEP-14 mentions `debian/latest` first, so I suspect many packages outside of the Go team will (or already do) use `debian/latest`.  It would be nice if all repos were consistent, otherwise why bother with DEP-14 at all?

Also, DEP-14 says, "we recommend the `<vendor>/<suite>` scheme over `<vendor>/<codename>`".  If the Go team doesn't want to use `debian/latest`, I would expect dh-make-golang to use `debian/unstable` instead of `debian/sid`.  Changing this alone is probably not worthwhile, but if `upstream` is going to be renamed to `upstream/<something>` then it might be a good opportunity to align with the recommendation.

(FWIW, I think DEP-14 is too weak, and that it should instead take an unambiguous opinionated stance.  If the strict guidelines wouldn't work well for a particular package then that package's maintainer can choose to not conform with DEP-14.)

>   3. Despite your experience, the go-team has been using DEP-14 successfully for many years, and there is absolutely no benefit in this change, but a huge cost.

Some benefits:
  * consistency that tooling can depend on
  * less surprising behavior for potential maintainers that are expecting `debian/latest`

I'm not familiar with all of the Go team's tooling, so I'm curious to know why renaming `upstream/sid` to `upstream/latest` would be a "huge cost".  I would expect incremental adoption to be feasible option:  a new repo would use the new branch name, while an existing repo would continue to use the old branch name until someone felt like renaming it.

(Renaming a published branch causes disruption in existing Git clones, but I think it's reasonable to assume that package maintainers are competent Git users and can easily deal with the disruption.  If that's too generous, a script can help with the tricky bits.)

ottok:
> Despite your experience, the go-team has been using DEP-14 successfully for many years, and there is absolutely no benefit in this change, but a huge cost.

Yes this is intentionally a suggestion for the Go team to start using `debian/latest` and change comes with some cost, but it is _not_ huge. New packages will have it automatically at no cost, and old packages can switch if they want by running [dep-14-convert-git-branch-names](https://manpages.debian.org/trixie/devscripts/dep-14-convert-git-branch-names.1.en.html). As long as the `debian-latest` value is correct in the `debian/gbp.conf` file nobody needs to do anything extra in daily packaging work, the usual commands like `gbp import-orig --uscan` will automatically work correctly.

n-peugnet:
> Yes this is intentionally a suggestion for the Go team to start using `debian/latest` and change comes with some cost, but it is _not_ huge.

I agree with @NightTsarina that the cost is huge, for almost no benefit. This will introduce even more fragmentation between the repositories of the team. IMO in the team,  package consistency is much more important than following latest recommendations. And introducing a new change will never replace the previous style on all packages, so it will only create more and more fragmentation.

IMO for such a disruptive change, it has to have a very obvious advantage. I still fail to see the concrete advantage this change provides.

> New packages will have it automatically at no cost, and old packages can switch if they want by running [dep-14-convert-git-branch-names](https://manpages.debian.org/trixie/devscripts/dep-14-convert-git-branch-names.1.en.html).

It is not just this that needs to be updated, it is all the repos cloned by everybody everywhere. Once again, for almost no benefit.

> As long as the `debian-latest` value is correct in the `debian/gbp.conf` file nobody needs to do anything extra in daily packaging work, the usual commands like `gbp import-orig --uscan` will automatically work correctly.

There will still be some places where one would have to know the branch and will have to check what is the layout used by a package, like basic git checkout.

rhansen:
> I agree with @NightTsarina that the cost is huge, for almost no benefit. This will introduce even more fragmentation between the repositories of the team. IMO in the team, package consistency is much more important than following latest recommendations.

If the latest recommendations can't be followed out of concern for inconsistency, then they're not recommendations—they're "wouldn't it be nice if" wishful thinking.

Any procedure change is going to cause some inconsistency.  If inconsistency is strongly discouraged, then it becomes very difficult to make any procedure changes.  Without procedure changes we can't make important improvements.

I think it's better for the Debian Go team to just accept some inconsistency as the cost of making progress.  Our tools will have to be written to accommodate the inconsistency.  When the inconsistency is minor (like this PR), we humans are smart enough to deal with it, and modify our tools to deal with it if necessary.  When it's major, we can create new tools to automate bulk changes across all projects, or at least flag the problems for manual correction.

As @ottok says, the costs are *not* high.  We already have to deal with different branch names in different repositories, so I don't believe the inconsistency argument.  And there's no need to update existing repos, so that's not a cost either.

> I still fail to see the concrete advantage this change provides.

Some benefits:

  * consistency (with the rest of Debian) that tooling can depend on
  * less surprising behavior for potential maintainers that are expecting `debian/latest`

ottok:
> > Yes this is intentionally a suggestion for the Go team to start using `debian/latest` and change comes with some cost, but it is _not_ huge.
> 
> I agree with @NightTsarina that the cost is huge, for almost no benefit. This will introduce even more fragmentation between the repositories of the team. IMO in the team, package consistency is much more important than following latest recommendations. And introducing a new change will never replace the previous style on all packages, so it will only create more and more fragmentation.

There is already fragmentation both inside the Go team, and between Go team and other team packages in this regard. Using `debian/latest` on new packages and slowly renaming old ones would eventually make the state uniform.
 
> IMO for such a disruptive change, it has to have a very obvious advantage. I still fail to see the concrete advantage this change provides.

Avoiding fragmentation is  a huge advantage, and also a lot of people misuse `debian/sid' vs `debian/experimental` without properly updating references when switching between them, which just using `debian/latest` by default totally solves (and that is probably why DEP-14 ended up recommending this).

> > New packages will have it automatically at no cost, and old packages can switch if they want by running [dep-14-convert-git-branch-names](https://manpages.debian.org/trixie/devscripts/dep-14-convert-git-branch-names.1.en.html).
> 
> It is not just this that needs to be updated, it is all the repos cloned by everybody everywhere. Once again, for almost no benefit.

Actually when you run `git pull` or `gbp pull` and your git version is recent enough the rename will happen automatically on your local checkouts after branches got renamed remotely.
 
> > As long as the `debian-latest` value is correct in the `debian/gbp.conf` file nobody needs to do anything extra in daily packaging work, the usual commands like `gbp import-orig --uscan` will automatically work correctly.
> 
> There will still be some places where one would have to know the branch and will have to check what is the layout used by a package, like basic git checkout.

Whenever you run `git clone` the default branch will be checked out automatically, and even better if running `gbp clone` and `debian/gbp.conf` having correct values all branches will end up being fetched correctly automatically.

A lot of repositories in Debian already renamed from `master` to `debian/latest` and it has gone mostly smoothly. You can test running [dep-14-convert-git-branch-names](https://manpages.debian.org/trixie/devscripts/dep-14-convert-git-branch-names.1.en.html) on some repo of yours and see that it won't cause much hassle to your collaborators.

rhansen:
> There is already fragmentation both inside the Go team, and between Go team and other team packages in this regard.

+1

> Actually when you run `git pull` or `gbp pull` and your git version is recent enough the rename will happen automatically on your local checkouts after branches got renamed remotely.

AFAIK, Git does not fetch the reflog (or similar) from remotes, so it doesn't have a way to reliably distinguish a renamed remote branch from a deleted+added pair of unrelated branches.  Do you have a reference?

In all versions of Git I'm familiar with, the upstream branch ("remote tracking branch"), such as `refs/remotes/origin/master`, is kind-of renamed if you do `git remote update --prune` (technically, Git deletes the old upstream branch and adds a new one).  But a local branch, such as `refs/heads/master`, is not automatically renamed to match.  Also, the local branch's associated upstream branch is not updated, so it continues to reference the now-gone upstream branch.

So renaming a published branch is disruptive to anyone that has already cloned the repository.  Assuming the local repo doesn't have any unpushed commits, this is easy to work around—just delete and re-clone.  If there are unpushed commits, the dev has to take extra steps to keep the commits.

But as pointed out before, the branches for existing repos don't have to be renamed.  Eventually it would be nice to rename them, and I think the benefit of consistency outweighs the potential for disruption.

Is https://flight-deck.debian.net/ something that could help us do bulk renames if desired?

ottok:
For reference, with git 2.43.0 you get this output when pulling a Go team repo with new branches, and a single `prune` is enough to fix references:

```
± git pull -fv
remote: Enumerating objects: 6, done.
remote: Counting objects: 100% (6/6), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 6 (delta 0), reused 0 (delta 0), pack-reused 0 (from 0)
Unpacking objects: 100% (6/6), 1.88 KiB | 482.00 KiB/s, done.
From salsa.debian.org:go-team/packages/golang-github-marusama-semaphore
   37abdc3..ee6910c  debian/sid      -> origin/debian/sid
 * [new branch]      debian/latest   -> origin/debian/latest
 * [new branch]      pristine-tar    -> origin/pristine-tar
error: cannot lock ref 'refs/remotes/origin/upstream/latest': 'refs/remotes/origin/upstream' exists; cannot create 'refs/remotes/origin/upstream/latest'
 ! [new branch]      upstream/latest -> origin/upstream/latest  (unable to update local ref)
error: some local refs could not be updated; try running
 'git remote prune origin' to remove any old, conflicting branches

± git remote prune origin
Pruning origin
URL: git@salsa.debian.org:go-team/packages/golang-github-marusama-semaphore.git
 * [pruned] origin/upstream
```

I will do another test with git 2.50 which has further improved automatic remote reference updates. I don't think that this migration has a "huge cost" and I believe long-term benefits outweigh the one-off cost of standardization work.

ottok:
Here a test with git 2.50 on a checkout of a repository where someone else did the rename:

```
± git pull -v
remote: Enumerating objects: 4, done.
remote: Counting objects: 100% (4/4), done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 4 (delta 0), reused 0 (delta 0), pack-reused 0 (from 0)
Unpacking objects: 100% (4/4), 1.97 KiB | 335.00 KiB/s, done.
From salsa.debian.org:go-team/packages/golang-github-canonical-go-efilib
 = [up to date]      debian/sid      -> origin/debian/sid
 * [new branch]      debian/latest   -> origin/debian/latest
 * [new branch]      upstream/latest -> origin/upstream/latest
error: some local refs could not be updated; try running
 'git remote prune origin' to remove any old, conflicting branches

± git remote prune origin
Pruning origin
URL: git@salsa.debian.org:go-team/packages/golang-github-canonical-go-efilib.git
 * [pruned] origin/upstream

± gbp pull --track-missing 
gbp:info: Fetching from default remote for each branch
gbp:error: Error running git branch: fatal: cannot lock ref 'refs/heads/upstream/latest': 'refs/heads/upstream' exists; cannot create 'refs/heads/upstream/latest'

± git branch -d upstream 
Deleted branch upstream (was 2af1bd9).

± gbp pull --track-missing 
gbp:info: Fetching from default remote for each branch
gbp:info: Branch 'upstream/latest' is already up to date.
gbp:info: Branch 'debian/latest' is already up to date.
```

So basically the key here is that collaborators need to run `git branch -d upstream` and/or `git pull --prune` to get the local `upstream` removed and replaced with the new `upstream/latest`.

PR #248, opened 2025-02-19
Stop modifying upstream .gitignore file and put build artifacts in debian/build/

ottok:
See commits for details. This is a re-submission of #230, pending to be merged potentially in the summer of 2025.

### DIFF
--- a/template.go
+++ b/template.go
@@ -86,6 +86,7 @@ func writeDebianGitIgnore(dir, debLib, debProg string, pkgType packageType) erro
 	fmt.Fprintf(f, "*.log\n")
 	fmt.Fprintf(f, "*.substvars\n")
 	fmt.Fprintf(f, "/.debhelper/\n")
+	fmt.Fprintf(f, "/build/\n")
 	fmt.Fprintf(f, "/debhelper-build-stamp\n")
 	fmt.Fprintf(f, "/files\n")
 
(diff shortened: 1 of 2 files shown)


template.go
```diff
@@ -299,7 +300,12 @@ func writeDebianRules(dir string, pkgType packageType) error {
 	fmt.Fprintf(f, "#!/usr/bin/make -f\n")
 	fmt.Fprintf(f, "\n")
 	fmt.Fprintf(f, "%%:\n")
-	fmt.Fprintf(f, "\tdh $@ --builddirectory=_build --buildsystem=golang\n")
+	fmt.Fprintf(f, "\tdh $@ --builddirectory=debian/build --buildsystem=golang\n")
+	// Note: The above `--builddirectory=debian/build` will eventually be obsolete
+	// in 2028+ then the dh-golang version 1.63+ that has
+	// https://salsa.debian.org/go-team/packages/dh-golang/-/commit/bc16dff5381b668a71fa99c381baba202c34c789
+	// is in use everywhere
+
```
NightTsarina, line 308: Considering `dh-golang` 1.63 is already in testing, I think we could remove this completely. After all, these tools are meant to be used from sid..

NightTsarina (commented):
LGTM to removing changes to upstream `.gitignore`, but maybe we can remove the custom `builddirectory` too?

template.go
guillemj: The omission of the --builddirectory option is dependent on the debhelper compat level, see the dh-golang(1) man page for the rationale, and debhelper compat 14 is not yet stable.

guillemj:
> LGTM to removing changes to upstream `.gitignore`, but maybe we can remove the custom `builddirectory` too?

From the commit reverting the previous change:

```
While the general idea makes sense and would make things better, this
had some push back on the review and still got merged anyway. It also
conflicts with an old proposal that still is pending to be updated
and re-proposed to try to unify the build paths for all packages at the
dpkg and debhelper layer, via a new dpkg-build-api level. And doing a
migration twice seems wasteful.

Ref: https://lists.debian.org/debian-dpkg/2020/03/msg00003.html
```

So, I agree with the idea of the change, but probably not in this specific form.

ottok:
Rebased on latest master

ottok:
In discussion with other Go team members the path debian/_build was more popular, so switched to it now.

ottok:
> Note that the standardized dpkg build dir proposal would move the contents into subdirectories of either `debian/.build/`and `debian/_build/` such as `debian/_build/builddir/`, so simply moving the build directory under `debian/_build/` would be incompatible with that and require an additional change once the proposal is implemented in some form. See https://www.hadrons.org/~guillem/tmp/draft.std-build-paths-v2 for my current draft, although I still need to polish multiple things there, where I've not yet found the time to do that, sorry. :/

Yes I am aware and as you see from the previous version of this PR it was using `debian/.build/`. However you didn't publish any proposals for dpkg despite several reminders in person and by email, so the schedule for this change Debian-wide is unclear, and as you didn't even publish the proposal it is not known yet what it will land on.

Several people said they dislike `debian/.build/` and prefer `debian/_build/` instead, which incidentally is also the minimal change needed to current situation to avoid people's need to modify the `.gitignore` file.

We can later switch to `debian/.build/` or some other path that hasn't yet been spec'ed if/when such a Debian-wide change comes in the future.

PR #254, opened 2025-02-19
Remove NOTE about upstream git history being a new feature

guillemj:
This has been in place for years now. No need to include this in the output which is long enough already.


ottok (approved):
LGTM!

NightTsarina (approved):

PR #255, opened 2025-02-20
Update dependencies to match versions in sid

NightTsarina:


guillemj:
Ugh, what is it with the CI asking about payment limits having been exceeded!?

guillemj (approved):
LGTM

PR #271, opened 2025-04-16
Fix progress and log output of estimate command

n-peugnet:
The progress was not tracking the correct directory and the output of "go get" is now omitted, except if it returns an error.

### DIFF
--- a/estimate.go
+++ b/estimate.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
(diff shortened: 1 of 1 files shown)


n-peugnet:
Here is a before and after.

**Before:**
before

**After:**
after

NightTsarina (approved):
LGTM

ottok (approved):

creekorful:
Thank you very much for your contribution!

PR #272, opened 2025-04-16
estimate: do not print children of mismatched major mods

n-peugnet:
Simply log a message once about mismatched major versions in the log output, and do not print all of their children.

For packages with a lot of dependencies, the previous behaviour was producing a lot of noise. If a module is already packaged, even with a different major version, it will probably not miss a lot of dependencies.

### DIFF
--- a/estimate.go
+++ b/estimate.go
@@ -95,15 +95,22 @@ func otherVersions(mod string) (mods []string) {
 }
 
 // findOtherVersion search in m for potential other versions of the given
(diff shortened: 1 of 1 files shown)


n-peugnet:
I just rebased and pushed an additional commit that displays the name of the corresponding Debian package for approximate matches, along with a link to the tracker page of this package (on terminal emulators that supports them).

2025-05-05-173728_821x557_scrot

ottok:
No other comments, proceeding to merging this.

Issue #263, opened 2025-03-11
[FR] ignore windows-only go modules


werdahias:
```
$ dh-make-golang estimate github.com/arduino/arduino-cli
[...]
2025/03/11 00:55:41 Bringing github.com/arduino/arduino-cli to Debian requires packaging the following Go modules:
github.com/arduino/arduino-cli
 fortio.org/safecast
 github.com/arduino/go-paths-helper
 github.com/arduino/go-properties-orderedmap
   github.com/arduino/go-paths-helper (2)
 github.com/arduino/go-serial-utils
 github.com/arduino/go-timeutils
 github.com/arduino/go-win32-utils
 github.com/arduino/pluggable-discovery-protocol-handler/v2
   github.com/arduino/go-paths-helper (3)
   github.com/arduino/go-properties-orderedmap (2)
 github.com/arduino/pluggable-monitor-protocol-handler
 github.com/cmaglie/pb
 github.com/codeclysm/extract/v4
   github.com/arduino/go-paths-helper (4)
   github.com/niemeyer/pretty
 github.com/djherbis/buffer
 github.com/djherbis/nio/v3
   github.com/djherbis/buffer (2)
 github.com/go-git/go-git/v5	(github.com/go-git/go-git in Debian)
   github.com/go-git/gcfg
   github.com/go-git/go-git-fixtures/v4	(github.com/go-git/go-git-fixtures in Debian)
   github.com/Microsoft/go-winio
 github.com/gofrs/uuid/v5	(github.com/gofrs/uuid in Debian)
 github.com/marcinbor85/gohex
 go.bug.st/cleanup
 go.bug.st/downloader/v2
 go.bug.st/f
 go.bug.st/relaxed-semver
 go.bug.st/testifyjson
 github.com/Microsoft/go-winio (2)
 github.com/go-git/gcfg (2)
 github.com/sagikazarmark/locafero
 github.com/sagikazarmark/slog-shim
```
go-winio and go-win32-utils are windows-only deps. it'd be nice if dh-make-golang would ignore those.

n-peugnet:
Good suggestion. I am not sure what would be the best way to implement thus. Were you simply thinking of a blacklist? This would probably be enough.

I also thought that maybe I could try to make the estimate command use the build graph again (instead of the mod graph currently). If it respects the build tags and if the usage of these libraries is garded by such build tags, it would ignore them by default.

But from what I saw, at least in docker packages, the import of the windows only libraries were not always garded by build tags.

[...]

PR #274, opened 2025-04-18
estimate: ignore some modules based on a blocklist

n-peugnet:
This is more in line with the work involved to package a module for Debian.

Closes: #263


ottok (approved):
Having a blocklist makes sense, and adding some Windows-only modules there initially to fix #263 is a good start. Thanks!

creekorful:
Thank you for your contribution! Sadly there isn't any other reliable way to filter those windows modules as of today.

This blocklist is indeed the easiest implementation.

PR #275, opened 2025-04-22
make: MIME-encode sender name when necessary

supertassu:
Otherwise sending the generated ITP files as is will fail for people
(like me) who have non-ASCII characters in their names.

### DIFF
--- a/make.go
+++ b/make.go
@@ -690,9 +690,11 @@ func writeITP(gopkg, debsrc, debversion string) (string, error) {
 		description = "TODO"
 	}
 
+	subject := mime.QEncoding.Encode("utf-8", fmt.Sprintf("ITP: %s -- %s", debsrc, description))
+
 	fmt.Fprintf(f, "From: %q <%s>\n", mime.QEncoding.Encode("utf-8", getDebianName()), getDebianEmail())
 	fmt.Fprintf(f, "To: submit@bugs.debian.org\n")
-	fmt.Fprintf(f, "Subject: ITP: %s -- %s\n", debsrc, description)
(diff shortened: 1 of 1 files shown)


make.go
```diff
@@ -689,7 +690,7 @@ func writeITP(gopkg, debsrc, debversion string) (string, error) {
 		description = "TODO"
 	}
 
-	fmt.Fprintf(f, "From: %q <%s>\n", getDebianName(), getDebianEmail())
+	fmt.Fprintf(f, "From: %q <%s>\n", mime.QEncoding.Encode("utf-8", getDebianName()), getDebianEmail())
 	fmt.Fprintf(f, "To: submit@bugs.debian.org\n")
 	fmt.Fprintf(f, "Subject: ITP: %s -- %s\n", debsrc, description)
```
creekorful, line 695: This could be impacted as well?

creekorful:
Hello, thank you very much for your contribution!

Since I have special characters in my name too, this PR is greatly appreciated :-)

I think it would be good to update `Subject` and `Owner` fields as well. What do you think?

make.go
supertassu: Fixed in 9374c6744f64831fbd14601c1cc6d8e686b61060.

supertassu:
> I think it would be good to update `Subject` and `Owner` fields as well. What do you think?

Those fields are in the email body, they're not real headers. So they're not affected as the code already sets `charset=utf-8` in the `Content-Type` header.

creekorful (approved):

creekorful:
Indeed you are right, only `Subject` needs to be updated.

Thank you for your contribution!

Issue #283, opened 2025-06-23
[feat] show packages in NEW


werdahias:
For cargo-debstatus we have a nice feature that shows dependencies in NEW in a light blue in the output, would be nice if d-m-g would do the same

PR #284, opened 2025-06-26
estimate: show packages in NEW, but with a light blue color

n-peugnet:
It uses an additional query from the DAK Web API to get the packages currently in NEW, but this one returns only source packages, so we have to keep into memory the name of the source packages when we get the list of all golang packages.

Maybe the name `getGolangBinaries` is a bit misleading now.

Closes: #283 

Here is an example output:
2025-06-27-182444_427x257_scrot

### DIFF
--- a/check_depends.go
+++ b/check_depends.go
@@ -91,7 +91,7 @@ func execCheckDepends(args []string) {
 // parseGoModDependencies parse ALL dependencies listed in go.mod
 // i.e. it returns the one defined in go.mod as well as the transitively ones
 // TODO: this may not be the best way of doing thing since it requires the package to be converted to go module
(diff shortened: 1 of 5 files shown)


n-peugnet:
Could maybe still improved but it is good for review IMO.
CC @werdahias

werdahias:
Thanks, looks very nice ! what's the best way to test this ?

n-peugnet:
> what's the best way to test this ?

I would say:

    git clone https://github.com/n-peugnet/dh-make-golang -b color-packages-in-new
    cd dh-make-golang
    go run . estimate <import-path>

ottok (approved):
This is a really nice addition and can help avoid a bunch of duplicate work

PR #290, opened 2025-08-12
Add dpkg-build-api (= 1) to the default build-deps

### LINKED ISSUE #289: Add `dpkg-build-api (= 1)` to the default list of build dependencies
Hi!

Commit <https://github.com/Debian/dh-make-golang/commit/0cbf0f067b67183220c168fd593d10158b633630> removed `Rules-Requires-Root: no` from the default list of build-deps, since this is now the default in trixie. Great!

Still, I think that instead of just removing it, we could do better. The value of `Rules-Requires-Root` is controlled by the dpkg build API version used by the build, and when unspecified this defaults to v0. Before `R-R-R` with API v0 switched from `binary-targets` to `no`, it would already be set to `no` under API v1.

Since using API v1 is recommended, but having to remember to add it manually for each package is a bit of a chore, what about adding it to the default set of build-deps, just like `debhelper-compat`?

You can find more documentation about it in [Policy § 5.6.31](https://www.debian.org/doc/debian-policy/ch-controlfields.html#rules-requires-root) and the [dpkg-build-api.7](https://manpages.debian.org/testing/dpkg-build-api.7) manpage.

Got this idea while writing this review comment: <https://salsa.debian.org/go-team/packages/golang-github-gonvenience-idem/-/merge_requests/2#note_638520>

Tachi107:
Closes: https://github.com/Debian/dh-make-golang/issues/289

### DIFF
--- a/check_depends_test.go
+++ b/check_depends_test.go
@@ -18,6 +18,7 @@ Priority: optional
 Build-Depends:
  debhelper-compat (= 13),
  dh-sequence-golang,
+ dpkg-build-api (= 1),
  golang-any,
  golang-github-advancedlogic-goose-dev,
  golang-github-fatih-color-dev,
--- a/template.go
+++ b/template.go
@@ -199,6 +199,7 @@ func writeDebianControl(dir, gopkg, debsrc, debLib, debProg string, pkgType pack
 	builddeps := append([]string{
 		"debhelper-compat (= 13)",
 		"dh-sequence-golang",
+		"dpkg-build-api (= 1)",
 		"golang-any"},
 		dependencies...)
 	sort.Strings(builddeps)


guillemj:
Thanks for the MR.  I think this is good in general, but I'm not sure whether the golang packages will currently take much benefit from specifying the `dpkg-build-api`, given that at least two of the behavior changes for v1 are now the default, and the others might not apply to golang packages in the common case. My concern would be with introducing unnecessary churn.

In any case, this would be  a +1 from me (in gerrit terms, as in "LGTM but someone else needs to approve with +2") given that I'm clearly biased here (from an implementer PoV), but if others think this is good and the churn is not a major concern, then let's go for it?

ottok (approved):
Motivations for this change seem reasonable after reading policy and dpkg-build-api man page as mentioned in #289. Approving, so it has +1 from me.

@guillemj Do you want to try approve as well so we can mimick the +1+2 functionality you longer from gerrit?

nileshpatra:
I just noticed `dpkg-build-api` this in a new go package and felt this was unnecessary. The then-fallouts in https://github.com/Debian/dh-make-golang/issues/289 are now fixed and by default work so I am unsure if adding this by default adds much benefit.

Could we maybe revisit this?

PS: I am not objecting to this, but just asking if this is really of much benefit.

Tachi107:
Hi Nilesh,

I think all the reasons for including it then also apply now. The fact 
that R³ is now a default does not change that (also: it was the default 
back then).

PR #291, opened 2025-08-25
Drop Priority: optional, which is the default in Trixie

elboulangero:
When omitted, the priority defaults to optional (since dpkg 1.22.13).

Cf. `man deb-src-control`.

### DIFF
--- a/check_depends_test.go
+++ b/check_depends_test.go
@@ -14,7 +14,6 @@ Uploaders:
  Aloïs Micard <alois@micard.lu>,
 Section: news
 Testsuite: autopkgtest-pkg-go
-Priority: optional
 Build-Depends:
  debhelper-compat (= 13),
  dh-sequence-golang,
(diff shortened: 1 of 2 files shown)


ottok (approved):
Thanks, we should indeed do this!

elboulangero:
Let's hold on a bit for this one. I tried, and noticed that Lintian complains:

```
Running lintian...
W: golang-github-oschwald-maxminddb-golang-v2 source: recommended-field (in section for source) Priority [debian/control:1]
```

```
$ lintian-explain-tags recommended-field
N:
W: recommended-field
N: 
N:   The named field is recommended by policy but not present in the package's primary control file.
N: 
N:   Please refer to Source package control files – debian/control (Section 5.2) in the Debian Policy Manual, Binary package control files – DEBIAN/control (Section 5.3) in the Debian Policy Manual, and Debian changes files – .changes (Section
N:   5.5) in the Debian Policy Manual for details.
N: 
N:   Visibility: warning
N:   Show-Always: no
N:   Check: fields/recommended
N:   Renamed from: no-priority-field no-section-field-for-source no-section-field no-urgency-in-changes-file
N: 
```

Indeed, the Debian policy was not yet updated to reflect this change: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1110017

elboulangero:
I tried to build a package without `Priority: optional`: no warning from Lintian. I think it can be merged now.

Issue #292, opened 2025-08-26
Out of memory with latest development version of `dh-make-golang estimate`


ottok:
When running `gh-make-golang estimate` on a large package, the process eats up a lot of memory and CPU, and eventually dies as OS kills it.

Steps to reproduce:

```
$± curl -LO https://salsa.debian.org/otto/dh-make-golang/-/jobs/8033558/artifacts/raw/debian/output/dh-make-golang_0.8+git20250820.a6fe7fa-1~otto+salsaci+20250820+39_amd64.deb
$ apt install ./dh-make-golang*.deb
$ dh-make-golang estimate github.com/ethereum/go-ethereum
2025/08/21 02:56:02 github.com/Azure/azure-sdk-for-go/sdk/storage/azblob is packaged as github.com/Azure/azure-sdk-for-go in Debian (golang-github-azure-azure-sdk-for-go)
2025/08/21 02:56:02 github.com/aws/aws-sdk-go-v2/config is packaged as github.com/aws/aws-sdk-go-v2 in Debian (golang-github-aws-aws-sdk-go-v2)
2025/08/21 02:56:02 github.com/aws/aws-sdk-go-v2/credentials is packaged as github.com/aws/aws-sdk-go-v2 in Debian (golang-github-aws-aws-sdk-go-v2)
2025/08/21 02:56:02 github.com/aws/aws-sdk-go-v2/service/route53 is packaged as github.com/aws/aws-sdk-go-v2 in Debian (golang-github-aws-aws-sdk-go-v2)
2025/08/21 02:56:02 github.com/cespare/xxhash/v2 has no version string in Debian (golang-github-cespare-xxhash)
2025/08/21 02:56:02 github.com/deckarep/golang-set/v2 has no version string in Debian (golang-github-deckarep-golang-set)
2025/08/21 02:56:02 github.com/golang-jwt/jwt/v4 has no version string in Debian (golang-github-golang-jwt-jwt)
2025/08/21 02:56:02 github.com/go-chi/chi/v5 has no version string in Debian (golang-github-go-chi-chi)
2025/08/21 02:56:02 github.com/labstack/echo/v4 has no version string in Debian (golang-github-labstack-echo)
2025/08/21 02:56:02 github.com/pion/stun/v2 has no version string in Debian (golang-github-pion-stun)
Killed
(out of memory)
```

Using the latest released version does not have this issue:
```
```
# dh-make-golang estimate github.com/ethereum/go-ethereum
[...]

PR #293, opened 2025-09-14
estimate: fix occasional infinite recursion

n-peugnet:
The "go mod graph" command returns a DAG so there can be no cycles, but modules can appear multiple times with different versions. When stripping the version from the modules we can accidentaly introduce cycles in the DAG by merging multiple versions of a module.

To prevent this, we go back to saving the exact output of "go mod graph" in memory, including the version strings, and get rid of them when visiting each node.

Closes: #292

### DIFF
--- a/estimate.go
+++ b/estimate.go
@@ -277,27 +277,25 @@ func estimate(importpath, revision string) error {
 		// imported it, separated by a single space. The module names
 		// can have a version information delimited by the @ character
 		src, dep, _ := strings.Cut(line, " ")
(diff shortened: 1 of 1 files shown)


ottok:
Nice!

ottok:
Yes I know how to reproduce the bug, I reported it :)

But saying that I haven't tested I meant that I didn't have time to test this yet. I did it now and I confirm it works.

image

ottok (approved):

PR #294, opened 2025-09-23
estimate: fix for modules that don't have a root package

n-peugnet:
Some modules don't have any .go files in their root directory (e.g. github.com/autobrr/qui), which means they have no root package. This led to the error "no required module provides package foo" when trying to list the direct dependencies of the estimated module.

To fix this, we now make sure to use "go list" in module mode using the "-m" option when looking for the module directory in the GOPATH.

EDIT: to make it clear, here is the error fixed by this change:

```console
$ dh-make-golang estimate github.com/autobrr/qui
no required module provides package github.com/autobrr/qui; to add it:
	go get github.com/autobrr/qui
2025/09/23 13:44:54 estimate: get direct dependencies: get module dir: go list: args: [go list -f {{.Dir}} github.com/autobrr/qui]; error: exit status 1
```

### DIFF
--- a/estimate.go
+++ b/estimate.go
@@ -110,6 +110,12 @@ func getDirectDependencies(gopath, repodir, module string) (map[string]bool, err
 	if err != nil {
 		return nil, fmt.Errorf("get module dir: %w", err)
 	}
+	// We cannot use "go list -m ..." if there is no go.mod file, but
+	// such packages are usually quite old and have few dependencies,
+	// so we return a nil map without error.
+	if _, err := os.Stat(filepath.Join(dir, "go.mod")); err != nil {
+		return nil, nil
+	}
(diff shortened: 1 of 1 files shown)


ottok:
I assume the issue this fixes was introduced in a6fe7fa873b9544187ff8bdc18aa8b3acc8061dd?

I tested `dh-make-golang estimate github.com/autobrr/qui` with both v0.8.0-1 in Debian unstable and a new [development version with this patch applied](https://salsa.debian.org/otto/dh-make-golang/-/jobs/8339383/artifacts/raw/debian/output/dh-make-golang_0.8+git20250820.a6fe7fa-1~otto+salsaci+20250923+43_amd64.deb) and got this output:

image

Hence we can see that the error that was likely is on current git head is gone with this change applied, thus approving.

ottok (approved):

PR #296, opened 2025-10-04
Ignore `/vN` major version suffix when populating `Upstream-Name`

rhansen:


ottok:
Alternatively add inline comment on line 262 showing an example of the "wrong" filename and how stripping the suffix fixes it.

PR #298, opened 2026-01-17
Append +ds suffix to upstream version when repacking

### LINKED ISSUE #141: upstream tag/version is not aligned with uscan file
https://github.com/Debian/dh-make-golang/blob/master/template.go#L326 adds +ds1 to uversion. But the initial version/tag created by dh-make-golang doesn't have this suffix.

aviralgarg05:
### Description
This PR resolves the issue where the upstream version/tag used in the initial packaging was not aligned with the `debian/watch` file expectations when repacking is involved.

Previously, `dh-make-golang` would correctly configure `debian/watch` with `repacksuffix=+ds1` if it detected that vendor directories or Godeps needed to be removed. However, the initial package version (used for the `.orig.tar` name and the first `debian/changelog` entry) did not include this suffix, leading to a mismatch that caused issues for tools like `uscan` and `gbp`.

#### Changes:
- Modified [make.go](cci:7://file:///Users/aviralgarg/Everything/dh-make-golang/make.go:0:0-0:0) to append the `+ds1` suffix to the upstream version whenever repacking is detected (`len(u.vendorDirs) > 0 || u.hasGodeps`).
- This ensures that:
    - The generated orig tarball name follows the pattern `package_version+ds1.orig.tar.xz`.
    - The `debian/changelog` entry starts with `version+ds1-1`.
    - The `debian/watch` file (which expects `+ds1` for repacked tarballs) is perfectly aligned with the package state.

### Related Issue
Fixes #141

### Verification Performed
- **Unit Tests**: Ran `go test ./...` and confirmed all tests pass.
- **Build**: Successfully built the tool using `go build`.
- **Linting/Formatting**:
    - Validated code with `go vet ./...`.
    - Ensured compliance with project style using `gofmt -s`.
- **Logic Verification**: Confirmed that the `u.version` modification happens after the upstream source analysis but before the Git repository and template generation, ensuring the correct version propagates to all generated files.

### Checklist
- [x] Code follows the style of the project.
- [x] Tests have been run and pass.
- [x] No unrelated files were modified.
- [x] Branch pushed: `fix-issue-141-version-alignment`


aviralgarg05:
Thanks for the reviews! 

**Logic and Timing**
The reason for checking `len(u.vendorDirs) > 0 || u.hasGodeps` here is that [dh-make-golang](cci:7://file:///Users/aviralgarg/Everything/dh-make-golang/dh-make-golang:0:0-0:0) automatically populates the `Files-Excluded` field in the generated `debian/copyright` whenever these directories are found. Since the tool is responsible for creating these templates, it effectively initiates the repackaging process. Adding the suffix at this stage ensures that the `d/changelog`, `d/watch`, and `d/copyright` files are consistent from the start.

**Suffix Convention**
I will update the code to use the standard `+ds` suffix instead of `+ds1` to align with Lintian recommendations and common Debian practice.

**Commit Message**
I will also rebase and update the commit message to follow the project convention

n-peugnet:
I do not have any strong opinion, I was simply against changing something for no reason. In this case the [Lintian tag `anticipated-repack-count`](https://lintian.debian.org/tags/anticipated-repack-count.html) mentioned by @umlaeute is quite explicit, so I think we'd better follow its advice.

ottok:
After thinking about this and re-reading the comments above, I think we should not have the digit and thus probably the optimal is to have `+ds` as the suffix?

ottok:
I was planning to merge this now, but in addition to the actual commit it has an extra merge commit. Can you rebase on latest 'master' or do you want me to merge this manually?

guillemj:
I think using +ds1 as default is better, it's just one more digit, and makes any potential repackaging bump uniform. This does not need to be a mistake, it can well be due to not needing vendored  modules anymore for example, because they have been uploaded as separate packages.

guillemj:
It also maps to other similar version designators where we do not start them w/o any version, but with 1, such as +nmu1, or +b1 or similar, which means having a uniform generic pattern means less rules to have to explain or remember, and thus more consistent overall versioning behavior.

ottok:
As discussed above, there is always also a Debian revision, so having extra digit is not necessary. E.g. after version `1.0.0+ds-1` you will have `1.0.0+ds-2`. To me having `1.0.0+ds1-1` you will have `1.0.0+ds2-2` would be just cluttering.

guillemj:
They are different things, one is the upstream marker the other the packaging revision after a repack iteration the revision gets reset, from 1.0.0+ds1-1 you'd go into 1.0.0+ds2-1 (not -2). If I saw a revision skip I'd wonder what happened here, and I'd be prompted to try to check the BTS or git history, which in this case would be useless, and inconclusive.

This goes back to avoiding to use strange and non-uniform versioning schemes, which either need explanation or documentation in the changelog or whareve, instead of using the obvious. Optimizing the version scheme to avoid a single character, at the cost of making this non-uniform and unobvious seems like the wrong thing to be optimizing on.

PR #300, opened 2026-02-19
Update debian/watch generation to use v5 templates

ottok:
In devscripts since 2.25.19 uscan automatically discovers GitHub upstream tarballs when debian/upstream/metadata contains the `Archive: GitHub` and `Repository` fields.  Since dh-make-golang already writes this file and only supports GitHub-hosted projects, the watch file is now redundant.

Removing it avoids maintenance overhead and confusion for new maintainers who might otherwise believe the file is required.

This also prevents Lintian from nagging about watch file format v4.

### DIFF
--- a/template.go
+++ b/template.go
@@ -414,45 +414,38 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
 	}
 	defer f.Close()
 
-	filenamemanglePattern := `s%(?:.*?)?v?(\d[\d.]*)\.tar\.gz%@PACKAGE@-$1.tar.gz%`
-	uversionmanglePattern := `s/(\d)[_\.\-\+]?(RC|rc|pre|dev|beta|alpha)[.]?(\d*)$/$1~$2$3/`
+	// This whole section uses the uscan v5 format
+	fmt.Fprint(f, "Version: 5\n")
 
+	// If upstream has releases, instruct uscan to download latest release.
+	// Otherwise use upstream git HEAD.
 	if hasRelease {
-		log.Printf("Setting debian/watch to track release tarball")
-		fmt.Fprint(f, "version=4\n")
-		fmt.Fprint(f, `opts="filenamemangle=`+filenamemanglePattern+`,\`+"\n")
-		fmt.Fprint(f, `      uversionmangle=`+uversionmanglePattern)
-		if repack {
-			fmt.Fprint(f, `,\`+"\n")
-			fmt.Fprint(f, `      dversionmangle=s/\+ds\d*$//,repacksuffix=+ds1`)
-		}
-		fmt.Fprint(f, `" \`+"\n")
-		fmt.Fprintf(f, `  https://%s/%s/%s/tags .*/v?(\d\S*)\.tar\.gz debian`+"\n", host, owner, repo)
+		log.Printf("Setting debian/watch to fetch the GitHub release tarball")
+		fmt.Fprint(f, "Template: GitHub\n")
+		fmt.Fprintf(f, "Dist: https://github.com/%s/%s\n", owner, repo)
 	} else {
 		log.Printf("Setting debian/watch to track git HEAD")
-		fmt.Fprint(f, "version=4\n")
-		fmt.Fprint(f, `opts="mode=git, pgpmode=none`)
-		if repack {
-			fmt.Fprint(f, `,\`+"\n")
-			fmt.Fprint(f, `      dversionmangle=s/\+ds\d*$//,repacksuffix=+ds1`)
-		}
-		fmt.Fprint(f, `" \`+"\n")
-		fmt.Fprintf(f, `  https://%s/%s/%s.git \`+"\n", host, owner, repo)
-		fmt.Fprint(f, "  HEAD debian\n")
+		fmt.Fprint(f, "Mode: git\n")
+		fmt.Fprintf(f, "Source: https://%s/%s/%s.git\n", host, owner, repo)
+		fmt.Fprint(f, "Matching-Pattern: HEAD\n")
+		fmt.Fprint(f, "# Enable verifying git tags are signed if upstream uses them:\n")
+		fmt.Fprint(f, "#Pgp-Mode: gittag\n")
 
 		// Anticipate that upstream would eventually switch to tagged releases
-		fmt.Fprint(f, "\n")
+		fmt.Fprint(f, "#\n")
 		fmt.Fprint(f, "# Use the following when upstream starts to tag releases:\n")
 		fmt.Fprint(f, "#\n")
-		fmt.Fprint(f, "#version=4\n")
-		fmt.Fprint(f, `#opts="filenamemangle=`+filenamemanglePattern+`,\`+"\n")
-		fmt.Fprint(f, `#      uversionmangle=`+uversionmanglePattern)
-		if repack {
-			fmt.Fprint(f, `,\`+"\n")
-			fmt.Fprint(f, `#      dversionmangle=s/\+ds\d*$//,repacksuffix=+ds1`)
-		}
-		fmt.Fprint(f, `" \`+"\n")
-		fmt.Fprintf(f, `#  https://%s/%s/%s/tags .*/v?(\d\S*)\.tar\.gz debian`+"\n", host, owner, repo)
+		fmt.Fprint(f, "#Template: GitHub\n")
+		fmt.Fprintf(f, "#Dist: https://github.com/%s/%s\n", owner, repo)
+	}
+
+	// If upstream release was repacked, use suffix '+ds' to continue the Go
+	// team's historical convention despite uscan man page stating '+ds' as
+	// 'useless' option and preferring '+dfsg'.
+	if repack {
+		log.Printf("Setting debian/watch to repack with suffix +ds")
+		fmt.Fprint(f, "Repacksuffix: +ds\n")
+		fmt.Fprint(f, "Dversion-Mangle: auto\n")
 	}
 
 	return nil
--- a/version.go
+++ b/version.go
@@ -24,7 +24,8 @@ var (
 
 	// uversionPrereleaseRegexp checks for upstream pre-release
 	// so that '-' can be replaced with '~' in pkgVersionFromGit.
-	// To be kept in sync with the regexp portion of uversionmanglePattern in template.go
+	// To be kept in sync with the default regex in uscan:
+	// https://manpages.debian.org/unstable/devscripts/debian-watch.5.en.html#Uversion-Mangle:
 	uversionPrereleaseRegexp = regexp.MustCompile(`(\d)[_\.\-\+]?(RC|rc|pre|dev|beta|alpha)[.]?(\d*)$`)
 )
 


guilherme-puida:
I think this makes uscan misbehave when upstream does not tag releases. I built this branch locally and tried `dh-make-golang make -type l github.com/tscholl2/siec`. Running `uscan --report-status` after results in:

```
uscan info: Scan watch files in .
uscan info: Check debian/watch and debian/changelog in .
uscan info: Found debian/upstream/metadata instead of debian/watch, trying to read it
uscan info: package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61-1" (as seen in debian/changelog)
uscan info: package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61" (no epoch/revision)
uscan info: ./debian/changelog sets package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61"
uscan warn: In /tmp/K8Cfn3Gdka/watch no matching files for watch source
  https://api.github.com/repos/tscholl2/siec/git/matching-refs/tags/
uscan info: Process watch file at: /tmp/K8Cfn3Gdka/watch
    package = golang-github-tscholl2-siec
    version = 0.0~git20240310.c2c6f61
    pkg_dir = .
uscan info: Parsing downloadurlmangle: s%(api.github.com/repos/[^/]+/[^/]+)/git/refs/%$1/tarball/refs/%g
uscan info: Parsing filenamemangle: s%.*/(?:[^/]+(?<=(?:\D|alpha|beta|rc))\-)?(?:[-_]?[Vv]?(\d[\-+\.:\~\da-zA-Z]*))%golang-github-tscholl2-siec-$1.tar.gz%
uscan info: Parsing pgpmode: none
uscan info: Parsing searchmode: plain
uscan info: Parsing versiontype: (?:[-_]?[Vv]?(\d[\-+\.:\~\da-zA-Z]*))
uscan info: Last orig.tar.* tarball version (from debian/changelog): 0.0~git20240310.c2c6f61
uscan info: Last orig.tar.* tarball version (dversionmangled): 0.0~git20240310.c2c6f61
uscan info: Requesting URL:
   https://api.github.com/repos/tscholl2/siec/git/matching-refs/tags/
uscan info: Scan finished
```

coldtobi:
does this change break tracker.d.o detection of new upstream versions?

ottok:
I switched the implementation to create a `debian/watch` file explicitly, but use v5 format. See commit for details.

Context:
- https://manpages.debian.org/unstable/devscripts/uscan-templates.5.en.html
- https://manpages.debian.org/unstable/devscripts/debian-watch.5.en.html
- https://manpages.debian.org/unstable/devscripts/uscan.1.en.html

template.go
```diff
@@ -414,45 +414,39 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
...
-		}
-		fmt.Fprint(f, `" \`+"\n")
-		fmt.Fprintf(f, `#  https://%s/%s/%s/tags .*/v?(\d\S*)\.tar\.gz debian`+"\n", host, owner, repo)
+		fmt.Fprint(f, "#Template: GitHub\n")
+		fmt.Fprintf(f, "#Dist: https://github.com/%s/%s\n", owner, repo)
+	}
+
+	// If upstream release was repacked, use suffix '+ds1' to continue the Go
+	// team's historical convention despite uscan man page stating '+dfsg' as
+	// primary option.
```
guilherme-puida: I think this comment is a little bit misleading: `+ds` and `+dfsg` serve different purposes and I don't think `+dfsg` is stated as a "primary option" anywhere.

template.go
```diff
@@ -414,45 +414,39 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
...
-		fmt.Fprint(f, `opts="filenamemangle=`+filenamemanglePattern+`,\`+"\n")
-		fmt.Fprint(f, `      uversionmangle=`+uversionmanglePattern)
-		if repack {
-			fmt.Fprint(f, `,\`+"\n")
-			fmt.Fprint(f, `      dversionmangle=s/\+ds\d*$//,repacksuffix=+ds1`)
-		}
-		fmt.Fprint(f, `" \`+"\n")
-		fmt.Fprintf(f, `  https://%s/%s/%s/tags .*/v?(\d\S*)\.tar\.gz debian`+"\n", host, owner, repo)
+		log.Printf("Setting debian/watch to fetch the GitHub release tarball")
+		fmt.Fprint(f, "Version: 5\n")
```
guilherme-puida: Duplicate `Version: 5` here.

template.go
```diff
@@ -414,45 +414,39 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
...
-		fmt.Fprintf(f, `  https://%s/%s/%s.git \`+"\n", host, owner, repo)
-		fmt.Fprint(f, "  HEAD debian\n")
+		fmt.Fprint(f, "Mode: git\n")
+		fmt.Fprintf(f, "Source: https://%s/%s/%s.git\n", host, owner, repo)
+		fmt.Fprint(f, "Matching-Pattern: HEAD\n")
+		fmt.Fprint(f, "# Enable verifying git tags are signed if upstream uses them:\n")
+		fmt.Fprint(f, "#Pgp-Mode: gittag\n")
 
 		// Anticipate that upstream would eventually switch to tagged releases
 		fmt.Fprint(f, "\n")
```
guilherme-puida: This breaks current uscan (`2.26.6`) AFAICT. With a `d/watch` file like this:

```
Version: 5
Mode: git
Source: https://github.com/tscholl2/siec.git
Matching-Pattern: HEAD
# Enable verifying git tags are signed if upstream uses them:
#Pgp-Mode: gittag

# Use the following when upstream starts to tag releases:
#
#Template: GitHub
#Dist: https://github.com/tscholl2/siec
```

This is the output of `uscan --report-status`

```
uscan info: Scan watch files in .
uscan info: Check debian/watch and debian/changelog in .
uscan info: package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61-1" (as seen in debian/changelog)
uscan info: package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61" (no epoch/revision)
uscan info: ./debian/changelog sets package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61"
uscan info: Process watch file at: debian/watch
    package = golang-github-tscholl2-siec
    version = 0.0~git20240310.c2c6f61
    pkg_dir = .
uscan warn: The following paragraph isn't well formatted, skipping it: << ==EOF==
# Use the following when upstream starts to tag releases:
#
#Template: GitHub
#Dist: https://github.com/tscholl2/siec
==EOF==
uscan info: Scan finished
```

If we remove the empty line it works fine:

```
uscan info: Scan watch files in .
uscan info: Check debian/watch and debian/changelog in .
uscan info: package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61-1" (as seen in debian/changelog)
uscan info: package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61" (no epoch/revision)
uscan info: ./debian/changelog sets package="golang-github-tscholl2-siec" version="0.0~git20240310.c2c6f61"
uscan info: Process watch file at: debian/watch
    package = golang-github-tscholl2-siec
    version = 0.0~git20240310.c2c6f61
    pkg_dir = .
uscan info: Parsing mode: git
uscan info: Last orig.tar.* tarball version (from debian/changelog): 0.0~git20240310.c2c6f61
uscan info: Last orig.tar.* tarball version (dversionmangled): 0.0~git20240310.c2c6f61
Cloning into bare repository '../golang-github-tscholl2-siec-temporary.6087.git'...
remote: Enumerating objects: 37, done.
remote: Counting objects: 100% (37/37), done.
remote: Compressing objects: 100% (35/35), done.
remote: Total 37 (delta 5), reused 13 (delta 2), pack-reused 0 (from 0)
Receiving objects: 100% (37/37), 66.94 KiB | 979.00 KiB/s, done.
Resolving deltas: 100% (5/5), done.
uscan info: uscan parser results:
    $base               = https://github.com/tscholl2/siec.git
    $filepattern        = HEAD
    $newfile            = HEAD
    $mangled_newversion = 0.0~git20240310.c2c6f61
    $newversion         = 0.0~git20240310.c2c6f61
    $lastversion        = 0.0~git20240310.c2c6f61
uscan info: Upstream URL(+tag) to download is identified as    https://github.com/tscholl2/siec.git HEAD
uscan info: Filename (filenamemangled) for downloaded file: golang-github-tscholl2-siec-0.0~git20240310.c2c6f61.tar.xz
uscan info: Newest version of golang-github-tscholl2-siec on remote site is 0.0~git20240310.c2c6f61, local version is 0.0~git20240310.c2c6f61
uscan info:  => Package is up to date from:
             => https://github.com/tscholl2/siec.git HEAD
uscan info: Removing git repo (../golang-github-tscholl2-siec-temporary.6087.git)
uscan info: Scan finished
```

template.go
n-peugnet: Here is the reasoning according to [debian-watch(5)](https://manpages.debian.org/trixie-backports/devscripts/debian-watch.5.en.html#Repacksuffix:):


> `+dfsg` is used when some non-free files are removed
> `+ds` is used when some files are removed for another reason (useless,...).
> `+repack` is used when source was re-downloaded without version number change

ottok: To me the man page stating that `+ds` is "useless" reads as `+dfsg` would be the primary option, right?

template.go
ottok: The `watch` file is supposed to support comments = ignore commented out lines, but it seems a bug in latest uscan v5: [Bug#1114473 uscan: spurious warning on commented-out paragraph](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1114473)

template.go
ottok: Thanks for noticing, dropped

template.go
ottok: Prefixed the newline now to satisfy this behavior.

template.go
ottok: Updated now the inline comment and also changed to `+ds` by default to be aligned with comments in #298 referring to both Lintian and uscan to default to not having a number suffix.

n-peugnet: > To me the man page stating that `+ds` is "useless" reads as `+dfsg` would be the primary option, right?

??
"useless" here is given as an example of "another reason" for removing some files.
Which is usually the reason why vendor directory are removed, and not because they are non-free. So +ds is definitely more appropriate than +dfsg in most cases for dh-make-golang.

template.go
```diff
@@ -414,45 +414,38 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
 	}
 	defer f.Close()
 
-	filenamemanglePattern := `s%(?:.*?)?v?(\d[\d.]*)\.tar\.gz%@PACKAGE@-$1.tar.gz%`
-	uversionmanglePattern := `s/(\d)[_\.\-\+]?(RC|rc|pre|dev|beta|alpha)[.]?(\d*)$/$1~$2$3/`
+	// This whole section uses the uscan v5 format
+	fmt.Fprint(f, "Version: 5\n")
```
guillemj, line 419: I'd like us to use the same format as documented in debian-watch(5), and shown in the examples, where `Version` is part of the first paragraph and the targets have their own paragraph (this also makes appending additional tarballs in MUT style packaging obvious).

template.go
```diff
@@ -414,45 +414,38 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
...
+		fmt.Fprint(f, "#Template: GitHub\n")
+		fmt.Fprintf(f, "#Dist: https://github.com/%s/%s\n", owner, repo)
+	}
+
+	// If upstream release was repacked, use suffix '+ds' to continue the Go
+	// team's historical convention despite uscan man page stating '+ds' as
+	// 'useless' option and preferring '+dfsg'.
+	if repack {
+		log.Printf("Setting debian/watch to repack with suffix +ds")
+		fmt.Fprint(f, "Repacksuffix: +ds\n")
```
guillemj: I'd prefer to use the more clear `Repack-Suffix` form for this. (I find it extremely unfortunate that the format considers `-` irrelevant and just strips them, because that means this file cannot be properly parsed by a standard deb822 parser, and makes it also harder to grep for.) (Even the debian-watch(5) documentation has a mix of this :( )

template.go
```diff
@@ -414,45 +414,38 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
...
+		fmt.Fprint(f, "Dversion-Mangle: auto\n")
```
guillemj: Nit: Personally I've got a slight preference for `DVersion-Mangle`, but it's not the standard way to "prettify" deb822 fields anyway (and this format is not a true deb822 one anyway). Although (not actionable here) I think ideally I'd prefer this to be named `Downstream-Version-Mangle` but this would all be https://bugs.debian.org/1113818. In any case this is just a nit, and I don't mind either way, just mentioning for your consideration.

template.go
```diff
@@ -414,45 +414,38 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
...
+		fmt.Fprint(f, "#Pgp-Mode: gittag\n")
 
 		// Anticipate that upstream would eventually switch to tagged releases
-		fmt.Fprint(f, "\n")
+		fmt.Fprint(f, "#\n")
 		fmt.Fprint(f, "# Use the following when upstream starts to tag releases:\n")
 		fmt.Fprint(f, "#\n")
-		fmt.Fprint(f, "#version=4\n")
-		fmt.Fprint(f, `#opts="filenamemangle=`+filenamemanglePattern+`,\`+"\n")
-		fmt.Fprint(f, `#      uversionmangle=`+uversionmanglePattern)
```
guillemj, line 447: These two options (filenamemangle and uversionmangle) get lost in the conversion. I agree we can rely on the default `Filename-Mangle` though. But we should add a `UVersion-Mangle` here.

template.go
```diff
@@ -414,45 +414,38 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
...
+	// This whole section uses the uscan v5 format
+	fmt.Fprint(f, "Version: 5\n")
 
+	// If upstream has releases, instruct uscan to download latest release.
+	// Otherwise use upstream git HEAD.
 	if hasRelease {
-		log.Printf("Setting debian/watch to track release tarball")
-		fmt.Fprint(f, "version=4\n")
-		fmt.Fprint(f, `opts="filenamemangle=`+filenamemanglePattern+`,\`+"\n")
-		fmt.Fprint(f, `      uversionmangle=`+uversionmanglePattern)
```
guillemj, line 422: These two options (filenamemangle and uversionmangle) get lost in the conversion. I agree we can rely on the default `Filename-Mangle` though. But we should add a `UVersion-Mangle` here.

template.go
```diff
@@ -414,45 +414,38 @@ func writeDebianWatch(dir, gopkg, debsrc string, hasRelease bool, repack bool) e
...
-		fmt.Fprintf(f, `#  https://%s/%s/%s/tags .*/v?(\d\S*)\.tar\.gz debian`+"\n", host, owner, repo)
+		fmt.Fprint(f, "#Template: GitHub\n")
+		fmt.Fprintf(f, "#Dist: https://github.com/%s/%s\n", owner, repo)
+	}
+
+	// If upstream release was repacked, use suffix '+ds' to continue the Go
+	// team's historical convention despite uscan man page stating '+ds' as
+	// 'useless' option and preferring '+dfsg'.
+	if repack {
+		log.Printf("Setting debian/watch to repack with suffix +ds")
```
guillemj: Switching from +ds1 to +ds is an unrelated change, which is covered in another MR, and where personally I don't agree with that change. Let's not entangle multiple unrelated things in the same MR, please. Also +dfsg and +ds serve different purposes, the former is for repacks due to (mainly) license issue, the latter for any kind of repackaging (such as stripping vendoring. Checking now the man page I read "useless" as referring to the reason for the files being in the tarball, not for the naming, but I think it's worth clarifying the wording in the man page as it seems very easy to misinterpret (I'll file a report or an MR).

guillemj (changes requested):
Thanks for preparing this! I had in mind doing this when I did a bunch of format conversions recently. :)

jas4711:
Thanks for working on this!  Looks good generally.

What do you think about not using GitHub release tarballs, even if they exist?

Instead track git directly.

For better supply-chain auditability with what upstream actually considers the sources they work on, I think relying on upstream git will be more reliable and easier to audit in the future.  Automatically generated tarballs may change hash (I think GitHub already made changes once or twice here), and maintainer-uploaded tarballs may contain xz-style concerns.

ottok:
> Thanks for working on this! Looks good generally.
> 
> What do you think about not using GitHub release tarballs, even if they exist?
> 
> Instead track git directly.
> 
> For better supply-chain auditability with what upstream actually considers the sources they work on, I think relying on upstream git will be more reliable and easier to audit in the future. Automatically generated tarballs may change hash (I think GitHub already made changes once or twice here), and maintainer-uploaded tarballs may contain xz-style concerns.

If you look at any of my Go packages or any other package I maintain (e.g. [ntfy](https://salsa.debian.org/go-team/packages/ntfy/-/network/debian%2Flatest), [entr](https://salsa.debian.org/debian/entr/-/network/debian%2Flatest)) or check my blog posts (e.g. [on creating new Debian packages](https://optimizedbyotto.com/post/debian-packaging-from-git/)) I always import upstream git when upstream has git available. Doing it makes the audit trail much better as showcased in e.g. my post on [XZ utils auditing](https://optimizedbyotto.com/post/xz-backdoor-debian-git-detection/). I know some people claim that using git would have prevented the XZ utils supply-chain attack, but that is not true as the attacker also had full control of upstream git and injected backdoor code in git as well, as explained in my blog post above. Debunking a myth and explaining what the attacker actually did or could have done in details is outside of the scope of a GitHub comment, so won't write more about that here now.

To have all Go packages import upstream git we should make the `includeUpstreamHistory` default in dh-make-golang, and update the Go team policy to advocate for proper use of `gbp.conf` with `upstream-vcs-tag` like I have been doing now for two years and been trying to drive consensus on. Before we discuss dropping support for tarballs completely I think we first need to have widespread adoption of importing upstreams via git first, and assurance that 99%+ of upstreams actually use git and include in their git releases everything that is needed for the Debian release (documentation etc). Until them we should continue migrating from tar-only to dual git+tar like I am using in all packages.

This PR about switching from uscan v4 to v5 does not include that in the scope but purely updates the watch file version. Jumping at this stage to remove tarball support I think is too big of a decision. The reliance on tarball checksums is broken in many ways as Debian does not actually save any checksums (e.g. sha256) in the package itself (e.g changelog), nor does Debian enforce that the import actually was done with a reproducible command (as seen in the stats at https://debaudit.debian.net/ only 60% of packages in Debian have identical tarball with the reproducible upstream tarball). The case about GitHub released tarballs changing paths and thus invalidating all old checksums isn't that bad as Debian stored upstream tarballs can be cross-referenced with the tarball checksums other distros have (and that is what e.g https://whatsrc.org/ is doing). I think the other reasons why orig tarballs are broken in Debian are way more severe. However I we need to fix the "political" issues in Debian preventing unified practices for supply-chain security before we do larger changes and try to migrate the ecosystem to something totally different and backwards-incompatible. In the mean time I think the best practice is what I am doing in my packages with uscan v5 + gbp upstreamvcs (which is also validated by having 100% suppy-chain pass rate on both upstream tarball and git in Debaudit for my packages).

jas4711:
> I know some people claim that using git would have prevented the XZ utils supply-chain attack, but that is not true as the attacker also had full control of upstream git and injected backdoor code in git as well, as explained in my blog post above. Debunking a myth and explaining what the attacker actually did or could have done in details is outside of the scope of a GitHub comment, so won't write more about that here now.

Agreed!  However I think the argument for git over tarball is cost: auditing the diff between git and the tarball has a cost, and few people do it.   Considering that it is often simple to switch to using git instead of tarballs, I think we should argue for doing that just because it lower the audit costs.

> To have all Go packages import upstream git we should make the `includeUpstreamHistory` default in dh-make-golang, and update the Go team policy to advocate for proper use of `gbp.conf` with `upstream-vcs-tag` like I have been doing now for two years and been trying to drive consensus on. Before we discuss dropping support for tarballs completely I think we first need to have widespread adoption of importing upstreams via git first, and assurance that 99%+ of upstreams actually use git and include in their git releases everything that is needed for the Debian release (documentation etc). Until them we should continue migrating from tar-only to dual git+tar like I am using in all packages.

I worry a bit about updating policy, since that often doesn't work well.  I'm not sure we'll be able to find proper wording that will meet consensus around the process (e.g., some people object to gbp), even though there could be consensus on the idea of using upstream git as the source.

> This PR about switching from uscan v4 to v5 does not include that in the scope but purely updates the watch file version. Jumping at this stage to remove tarball support I think is too big of a decision.

Indeed -- let's not prevent merging this waiting for any resolution on the git vs tarball question.  I feel a bit sorry for bringing it up, instead of just giving +1 on the PR.

/Simon

template.go
ottok: I updated the spelling to the variant you suggested and I also posted my +1 on Debian#1113818 to have uscan use more clear terminology (and spelling).

template.go
ottok: I don't understand the comment. The `UVersion-Mangle` is no needed here. Are you suggesting we add an commented out example line with `UVersion-Mangle` and a comment on when to add enable it fi needed?

ottok:
I am merging this now. Discussion on having git-only imports continues in #327.

PR #305, opened 2026-02-25
Relax acceptable `XS-Go-Import-Path` syntax

rhansen:
This PR relaxes how `XS-Go-Import-Path` values are parsed to improve maintainer friendliness:

  * Filter out empty string values in `XS-Go-Import-Path` (in particular, ignore a trailing comma).
  * Allow values to be separated by spaces and/or commas.

### DIFF
--- a/search.go
+++ b/search.go
@@ -32,7 +32,7 @@ type getGolangBinariesConfig struct {
 
 type getGolangBinariesOption func(cfg *getGolangBinariesConfig)
 
-func getGolangBinariesUrl(url string) getGolangBinariesOption {
+func withGolangBinariesUrl(url string) getGolangBinariesOption {
 	return func(cfg *getGolangBinariesConfig) {
 		cfg.url = url
 	}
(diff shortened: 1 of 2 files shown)


ottok:
Do we really need the field to support both using commas or space as separator?

```
    XS-Go-Import-Path:
     example.com/foo
     example.com/bar
```

In `debian/control` files most fields use either comma or space as separator, but never both. Allowing multiple formats sets the stage for having inconsistent formatting, which I am personally not a fan of.

rhansen:
> Do we really need the field to support both using commas or space as separator?

Do you mean a mix of space- and comma-separated items within a single `XS-Go-Import-Path` field?  For example:

```
XS-Go-Import-Path:
  example.com/foo
  example.com/bar,
  example.com/baz
```

If that's what you mean, then no, we don't need to support that.  But:
  * I don't expect a mix like that to happen in practice, at least not often enough to be a concern.
  * Asserting consistent separators within a particular `XS-Go-Import-Path` field adds code complexity.
  * We can add a lintian check to encourage consistency.

Maytha8:
In practise, I've only seen them comma-separated. I'm pretty sure that's how the ftp-master API parses them.

guillemj:
While I think I might have preferred to have this field be separated by spaces instead of commas, this seems too late now to switch, and not worth the potential fallout.

A brief check on dak, seems to imply it does not care about the shape of the field's value.

I also stumbled over this recently when making dh-golang more robust when parsing this field, and ended up making sure spaces are not considered separators, but final commas are accepted.

I don't know what else might be parsing this field, either.

PR #311, opened 2026-03-13
tests: Add unit tests for getGolangBinaries

rhansen:
This is one of the commits from #305 in its own PR because it's probably less controversial than the other commits.


ottok (approved):
Thanks! This package indeed did not have much tests. Using standard `go-cmp` to have these makes sense to me.

n-peugnet (approved):

PR #314, opened 2026-04-30
make: add -https flag to use HTTPS git remote instead of SSH

### LINKED ISSUE #199: Option to create HTTPS Git remote instead of SSH
`create-salsa-project` creates an SSH Git remote:

https://github.com/Debian/dh-make-golang/blob/52983fc8eecb1ab488d7efa28d0298b65317060a/make.go#L456

It would be neat to have an option to create an HTTPS remote instead

arnav03op:
Fixes #199

Added a -https flag to the make subcommand. When passed, the git
remote origin is set to an HTTPS URL instead of the default SSH URL.

SSH remains the default->>no breaking change.

Tested with:
- go build ./...
- go test ./...
- ./dh-make-golang make -help (flag appears correctly)

### DIFF
--- a/create_salsa_project.go
+++ b/create_salsa_project.go
@@ -13,9 +13,14 @@ import (
 func execCreateSalsaProject(args []string) {
 	fs := flag.NewFlagSet("create-salsa-project", flag.ExitOnError)
 
+	useHTTPS := fs.Bool("https", false, "Use HTTPS remote URL instead of SSH")
+
 	fs.Usage = func() {
-		fmt.Fprintf(os.Stderr, "Usage: %s create-salsa-project <project-name>\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "Usage: %s create-salsa-project [FLAG]... <project-name>\n", os.Args[0])
 		fmt.Fprintf(os.Stderr, "Example: %s create-salsa-project golang-github-mattn-go-sqlite3\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "\n")
+		fmt.Fprintf(os.Stderr, "Flags:\n")
+		fs.PrintDefaults()
 	}
 
 	if err := fs.Parse(args); err != nil {
@@ -44,4 +49,6 @@ func execCreateSalsaProject(args []string) {
 		b, _ := io.ReadAll(resp.Body)
 		log.Fatalf("unexpected HTTP status code: got %d, want %d (response: %s)", got, want, string(b))
 	}
+
+	fmt.Printf("Project created. Remote URL: %s\n", salsaPackageURL(projectName, *useHTTPS))
 }
--- a/make.go
+++ b/make.go
@@ -416,8 +416,15 @@ func runGitCommandIn(dir string, arg ...string) error {
 	return cmd.Run()
 }
 
+func salsaPackageURL(debsrc string, useHTTPS bool) string {
+	if useHTTPS {
+		return fmt.Sprintf("https://salsa.debian.org/go-team/packages/%s.git", debsrc)
+	}
+	return fmt.Sprintf("git@salsa.debian.org:go-team/packages/%s.git", debsrc)
+}
+
 func createGitRepository(debsrc, gopkg, orig string, u *upstream,
-	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string, pristineTar bool) (string, error) {
+	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string, pristineTar bool, useHTTPS bool) (string, error) {
 	wd, err := os.Getwd()
 	if err != nil {
 		return "", fmt.Errorf("get cwd: %w", err)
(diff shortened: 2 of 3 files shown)


arnav03op:
Fixed the issue!

arnav03op:
hey @ctrochalakis @olasd @gwolf @yarikoptic can you plz review the commit, if any changes required i will gladly look into it.

ottok:
@rhansen @n-peugnet Would you like to review this?

create_salsa_project.go
```diff
@@ -29,6 +33,8 @@ func execCreateSalsaProject(args []string) {
 
 	projectName := fs.Arg(0)
 
+	
+
```
guillemj: Spurious blank lines change.

create_salsa_project.go
```diff
@@ -13,9 +13,13 @@ import (
 func execCreateSalsaProject(args []string) {
 	fs := flag.NewFlagSet("create-salsa-project", flag.ExitOnError)
 
+	useHTTPS := fs.Bool("https", false, "Use HTTPS remote URL instead of SSH")
+
 	fs.Usage = func() {
 		fmt.Fprintf(os.Stderr, "Usage: %s create-salsa-project <project-name>\n", os.Args[0])
 		fmt.Fprintf(os.Stderr, "Example: %s create-salsa-project golang-github-mattn-go-sqlite3\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "\nFlags:\n")
```
guillemj: Print the blank line on its own, do not prefix it to the Flags. So that it's more clear what will happen on output, and also makes it easier to shuffle around. This follows existing practice in this project.

create_salsa_project.go
```diff
@@ -40,8 +46,21 @@ func execCreateSalsaProject(args []string) {
 	if err != nil {
 		log.Fatalf("http post: %s", err)
 	}
+
+	defer resp.Body.Close()
+
```
guillemj: Why this change? It looks like an unrelated fix that should at least be on its own commit, but maybe even on its own PR?

make.go
```diff
@@ -417,7 +417,7 @@ func runGitCommandIn(dir string, arg ...string) error {
 }
 
 func createGitRepository(debsrc, gopkg, orig string, u *upstream,
-	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string, pristineTar bool) (string, error) {
+	includeUpstreamHistory bool, allowUnknownHoster bool, debianBranch string, pristineTar bool,useHTTPS bool) (string, error) {
```
guillemj: Missing space after comma

make.go
```diff
@@ -449,7 +449,14 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
 
 	// [remote "origin"]
 
-	originURL := "git@salsa.debian.org:go-team/packages/" + debsrc + ".git"
+	// originURL := "git@salsa.debian.org:go-team/packages/" + debsrc + ".git"
```
guillemj: Do not leave around old code in commented form, just remove it.

make.go
```diff
@@ -449,7 +449,14 @@ func createGitRepository(debsrc, gopkg, orig string, u *upstream,
...
+	var originURL string
+	if useHTTPS {
+    	originURL = "https://salsa.debian.org/go-team/packages/" + debsrc + ".git"
+	} else {
+    	originURL = "git@salsa.debian.org:go-team/packages/" + debsrc + ".git"
```
guillemj: This should be refactored into its own function, so that we do not duplicate the logic with the other instance above, which also uses a different way to construct the string, which also seems weird.

make.go
```diff
@@ -960,7 +968,9 @@ func execMake(args []string, usage func()) {
 
 	debversion := u.version + "-1"
 
-	dir, err := createGitRepository(debsrc, gopkg, orig, u, includeUpstreamHistory, allowUnknownHoster, debBranch, pristineTar)
+	//dir, err := createGitRepository(debsrc, gopkg, orig, u, includeUpstreamHistory, allowUnknownHoster, debBranch, pristineTar,*useHTTPS)
```
guillemj: Same comment about not leaving old code around.

create_salsa_project.go
```diff
@@ -40,8 +46,21 @@ func execCreateSalsaProject(args []string) {
...
 		log.Fatalf("unexpected HTTP status code: got %d, want %d (response: %s)", got, want, string(b))
 	}
+
+	// Print the remote URL that will be used (informational)
+    var remoteURL string
+    if *useHTTPS {
+        remoteURL = fmt.Sprintf("https://salsa.debian.org/go-team/packages/%s.git", projectName)
+    } else {
+        remoteURL = fmt.Sprintf("git@salsa.debian.org:go-team/packages/%s.git", projectName)
+    }
```
guillemj: (See comment below about refactoring.)

guillemj (changes requested):

arnav03op:
Hi @guillemj , sorry for the delay , i'll make sure that all the changes will be done by today.

arnav03op:
@guillemj i made the changes you have asked for, plz review it

guillemj:
@arnav03op Thanks for the changes! Although why are the ones for create_salsa_project.go dropped now?

arnav03op:
Thanks for checking. I dropped the create_salsa_project.go changes because I thought they were too loosely related to the make command remote configuration, but I see now that this also removed part of the originally proposed behavior.
I’ll restore the create-salsa-project bits, but clean them up according to your earlier comments: remove the spurious blank lines, keep the usage formatting consistent, avoid the unrelated defer change, and reuse the same Salsa URL helper instead of duplicating URL construction.

arnav03op:
Thanks, you were right. I had dropped those while trying to keep the change scoped, but that also removed part of the original intended behavior.
I restored the create_salsa_project.go changes now, cleaned up according to your earlier comments, and reused the shared Salsa URL helper so the URL construction is not duplicated. I also left out the unrelated resp.Body.Close change.

PR #317, opened 2026-06-04
Stop creating a debian/.gitignore file

ottok:
Multiple people requested on the Go team mailing list that we stop with the practice of creating and maintaining .gitignore files. Instead each package should have appropriate `debian/clean` rules or maintainers should just ignore all extra artifacts created in a build without needing git to ignore them.

See thread https://lists.debian.org/debian-go/2026/05/msg00120.html for details.

### DIFF
--- a/template.go
+++ b/template.go
@@ -34,9 +34,6 @@ func writeTemplates(dir, gopkg, debsrc, debLib, debProg, debversion string,
 		return fmt.Errorf("mkdir debian/source/: %w", err)
 	}
 
-	if err := writeDebianGitIgnore(dir, debLib, debProg, pkgType); err != nil {
-		return fmt.Errorf("write debian/.gitignore: %w", err)
-	}
 	if err := writeDebianChangelog(dir, debsrc, debversion); err != nil {
 		return fmt.Errorf("write changelog: %w", err)
 	}
@@ -71,38 +68,6 @@ func writeTemplates(dir, gopkg, debsrc, debLib, debProg, debversion string,
 
 	if err := writeDebianGitLabCI(dir); err != nil {
 		return fmt.Errorf("write GitLab CI: %w", err)
-	}
-
-	return nil
-}
-
-func writeDebianGitIgnore(dir, debLib, debProg string, pkgType packageType) error {
-	f, err := os.Create(filepath.Join(dir, "debian", ".gitignore"))
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-
-	fmt.Fprintf(f, "*.debhelper\n")
-	fmt.Fprintf(f, "*.log\n")
-	fmt.Fprintf(f, "*.substvars\n")
-	fmt.Fprintf(f, "/.debhelper/\n")
-	fmt.Fprintf(f, "/_build/\n")
-	fmt.Fprintf(f, "/debhelper-build-stamp\n")
-	fmt.Fprintf(f, "/files\n")
-
-	switch pkgType {
-	case typeLibrary:
-		fmt.Fprintf(f, "/%s/\n", debLib)
-	case typeProgram:
-		fmt.Fprintf(f, "/%s/\n", debProg)
-	case typeLibraryProgram:
-		fallthrough
-	case typeProgramLibrary:
-		fmt.Fprintf(f, "/%s/\n", debLib)
-		fmt.Fprintf(f, "/%s/\n", debProg)
-	default:
-		log.Fatalf("Invalid pkgType %d in writeDebianGitIgnore(), aborting", pkgType)
 	}
 
 	return nil


nileshpatra (approved):

guillemj:
If people are annoyed by the .gitignore file, I'd really prefer if this could instead be disabled with an option, or if a majority don't want to see them generated I could live with that being the default (although the later would not be my preference at all).

I still prefer to not get my «git status» filled with unrelated files when working on my git work tree after a build, where you might end up otherwise «git add -A» unintended artifacts by accident, f.ex.

The `debian/clean` suggestions on the list do not seem relevant, or would be the wrong thing to do, as all these files that dpkg or debhelper generate and are aware of, will get deleted in the clean target (including whatever builddirectory was set to).

(IMO any tool that misbehaves due to .gitignore files being missing on archive creation, is misdesigned, but oh well… :D )

jas4711:
I still don't understand what kind of package build flow makes you notice any difference when having vs not having the normal Go team debian/.gitignore file?  I never notice any behavioural difference with or without that file.  I'm using 'gbp' to build packages, but understand that using 'sbuild' also does not trigger this.  Are people calling 'dpkg-buildpackage' manually, and then becoming confused about remaining build artifacts?

I don't think it is dgit misbehaving that is the concern here: the reason is that debian/.gitignore is incompatible with Debian source package <-> Debian git repository reproducibility/auditability.  That dgit does not support that is a feature of dgit, not a problem, IMHO.  The problem is having inconsistency between Debian source packages and git repo in the first place.  I do think dgit supports .gitignore differences with --quilt=gbp, but I find that a misfeature and avoid using it.

nileshpatra:
Since not everyone wants to drop `d/.gitignore`, would it be a fair compromise to still have it but have it disabled by default?

jas4711:
> Since not everyone wants to drop `d/.gitignore`, would it be a fair compromise to still have it but have it disabled by default?

Works for me!  Anything that moves us further towards source package reproducibility seems good to me.

guillemj:
> I still don't understand what kind of package build flow makes you notice any difference when having vs not having the normal Go team debian/.gitignore file? I never notice any behavioural difference with or without that file. I'm using 'gbp' to build packages, but understand that using 'sbuild' also does not trigger this. Are people calling 'dpkg-buildpackage' manually, and then becoming confused about remaining build artifacts?

I've mentioned parts of this in the mailing list. Using any tool that builds in-place will show this. This is the equivalent of doing an in-place/in-tree build for an upstream project. Personally I do run `dpkg-buildpackage` so that I can dog-food this and test it w/o any intermediate layers, also because all those layers add unnecessary complexity and points of failure. People do get confused, because I've seen auto-generated artifacts being committed. I personally just get annoyed because not having `debian/.gitignore` files increases friction when dealing with source repos.
 
> I don't think it is dgit misbehaving that is the concern here: the reason is that debian/.gitignore is incompatible with Debian source package <-> Debian git repository reproducibility/auditability. That dgit does not support that is a feature of dgit, not a problem, IMHO. The problem is having inconsistency between Debian source packages and git repo in the first place. I do think dgit supports .gitignore differences with --quilt=gbp, but I find that a misfeature and avoid using it.

All packages I maintain have appropriate `debian/.gitignore` files, and all of them are source package reproducible, because `dpkg-source` consistently excludes those files. The problem is that `dgit` does not follow this behavior, and thus it causes unreproducible output, this is a problem of its own making. One workaround for people that want to keep using `dgit`, is to configure the source package so that `dpkg-source` does not exclude those files, but that is currently cumbersome, because that requires re-filling all the other ignore values that are currently set by default. I've had pending making this easier to do, but other things have taken precedence, and this to me still seems like a workaround for a bad design descision.

ottok:
I find it somewhat contradicting that you Guillem Jover consider the use of a `debian/.gitignore` useful, but at the same time [you insist to keep the Go build artifacts in `_build`](https://salsa.debian.org/go-team/packages/dh-golang/-/merge_requests/26) instead of a path under `debian/` (e.g. `debian/_build`), thus rendering the use of a `debian/.gitignore` moot as it can't be used to tell git to ignore the build artifacts generated by the Go build.

We should choose either avenue and then be consistent with it.

jas4711:
> I find it somewhat contradicting that you Guillem Jover consider the use of a `debian/.gitignore` useful, but at the same time [you insist to keep the Go build artifacts in `_build`](https://salsa.debian.org/go-team/packages/dh-golang/-/merge_requests/26) instead of a path under `debian/` (e.g. `debian/_build`), thus rendering the use of a `debian/.gitignore` moot as it can't be used to tell git to ignore the build artifacts generated by the Go build.
> 
> We should choose either avenue and then be consistent with it.

IMHO a better avenue than chosing between those is to just drop both usages.  That is, stop creating `debian/.gitignore`, and stop adding `--builddirectory=_build`.  Lower complexity is better -- go with the defaults if we cannot justify a discrepency.

PR #318, opened 2026-06-04
Don't pass --buildsystem=golang to dh

jas4711:
According to Guillem Jover it is not needed with dh 13.4 or later.

### DIFF
--- a/template.go
+++ b/template.go
@@ -310,7 +310,7 @@ func writeDebianRules(dir string, pkgType packageType) error {
 	fmt.Fprintf(f, "#!/usr/bin/make -f\n")
 	fmt.Fprintf(f, "\n")
 	fmt.Fprintf(f, "%%:\n")
-	fmt.Fprintf(f, "\tdh $@ --builddirectory=debian/_build --buildsystem=golang\n")
+	fmt.Fprintf(f, "\tdh $@ --builddirectory=debian/_build\n")
 	// Note: The above `--builddirectory=debian/_build` will eventually be obsolete
 	// in 2028+ then the dh-golang version 1.64+ that has merged
 	// https://salsa.debian.org/go-team/packages/dh-golang/-/merge_requests/26


ottok:
"According to somebody" is not very convincing. Did you test on Debian Trixie removing that section from a package? Current oldstable has debhelper 13.11, so this has been obsolete for years already?

Should we also remove `--builddirectory=debian/_build` and let dh-golang defaults decide the build directory?

jas4711:
I've found the old bug report about this:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=981106

It talks about only working if there is a go.mod.  Do dh-make-golang work without go.mod?

I think it make sense to attempt to reduce the number of non-standard ways, and rely on default values, to as much extent as possible.  But not necessarily at the cost of breaking some packages, which would otherwise work.  Maybe it could add `--buildsystem=golang` only if there is no go.mod?  Assuming that comment is actually accurate.

Indeed more testing is necessary.  For example, do we care about generated `debian/rules` being usable for backports? oldstable? oldoldstable?  There ought to be some limit where we draw the line.

ottok:
This dh-make-golang is only for new packages. Totally new packages
would not go to old Debian releases, so we don't need to consider
them.

My question about old versions was just that the version you mentioned
(13.4) is very old and I was suprised that something would be so
obsolete yet not cleaned away from basically anywhere.

gibmat:
> This dh-make-golang is only for new packages. Totally new packages would not go to old Debian releases, so we don't need to consider them.

This is incorrect; backporting the Incus 7.0 LTS release for trixie required backporting a few new packages to a trixie environment. At a minimum any changes should be expected to work in trixie to support backporting work.

guillemj:
As @gibmat mentions this is not as old as it is claimed here, oldstable does not currently have a new enough debhelper version.

Not setting an explicit `--buildsystem` is going to be potentially confusing for (new) maintainers and error prone, and trying to make the detection in dh-make-golang match the logic in debhelper + dh-golang is going to be cumbersome to keep in sync (as it would need to reencode the dh buildsystem priority lists, and all their detection logic).

The current list of files that dh-golang uses to decide whether it can autobuild golang stuff is  `go.mod GLOCKFILE Godeps/Godeps.json Gopkg.lock glide.lock`. But if the source tree contains any of the files that debhelper uses to detect other build systems that have a higher priority (for example a `Makefile`), then golang will not be picked up, and instead we might end up building with make or other buildsystems (several of the prometheus packages are an example of this).

Removing that option should be done explicitly by a human, IMO. And I think we should not merge this one, so a -2 from me.

guillemj:
Maybe as a compromise for now, we could leave the current option as-is, but add a TODO comment before the "dh" invocation in `debian/rules` stating that one might be able to remove the `--buildsystem=golang` if after build and testing everything looks good, and if there are no other build systems that might have precedence.

In the future, say after oldstable stops being supported as LTS, the option could stop being emitted, and the TODO item be switched to remind the maintainer that might need to add it back in case the package mis-builds or misbehaves in general, so that at least there's an indication for what might be going wrong. Instead of leaving maintainers stumped about what's happening (after they supposedly did everything correctly).

jas4711:
I agree duplicating the dh logic for when to add `--buildsystems=golang` (wrt go.mod etc) does not seem like a good idea.

I don't follow why we should optimize future behaviour of dh-make-golang for oldstable.  Even optimizing it for stable, merely for the exceptional few backports of NEW packages that we may see, seems excessive.

Would maintainers back-porting future NEW packages into trixie really not know about adding `--buildsystem=golang`?  If that is really needed, it seems both stable and oldstable shipped with modern enough dh?

Maybe we need some experimentating -- take some package aimed at trixie-backports and remove the `--buildsystem=golang` and see if `dh` handle it.

jas4711:
> In the future, say after oldstable stops being supported as LTS, the option could stop being emitted, and the TODO item be switched to remind the maintainer that might need to add it back in case the package mis-builds or misbehaves in general, so that at least there's an indication for what might be going wrong. Instead of leaving maintainers stumped about what's happening (after they supposedly did everything correctly).

I think we could do that approach now.

ottok:
What is the conclusion here? To just merge this now as-is, and stop using `--buildsystem` on all new packages?

jas4711:
I don't have a strong opinion, but I'm looking for some stronger justification for the `--buildsystem golang`.  I've been removing it from packages I've uploaded lately, and mostly it has worked fine, but there has been some exceptions for rather old upstreams.  Maybe it is related to missing go.mod: understanding when not using this break things would be nice (I'll try to take notes next time it occurs).

ottok:
I tested removing this on a package. On Debhelper 13 the build failed on `dwz: debian/ntfy/usr/bin/ntfy: Unknown debugging section .debug_addr` and on Debhelper 14 build passes but I started seeing this warning:

```
dpkg-gencontrol: warning: Built-Using field of package ntfy: substitution variable ${misc:Built-Using} used, but is not defined
dpkg-gencontrol: warning: Static-Built-Using field of package ntfy: substitution variable ${misc:Static-Built-Using} used, but is not defined
```

ottok:
On some packages removing the `--buildsystem=golang` will start invoking a `Makefile` if the project has it, and one sees errors like:

```
   dh_auto_build
	make -j4 INSTALL="install --strip-program=true"
...
pkg/tuiv2/progress/tty_bubbletea.go:7:2: module lookup disabled by GOPROXY=off
pkg/tuiv2/progress/tty_bubbletea.go:8:2: module lookup disabled by GOPROXY=off
make[1]: *** [Makefile:110: vet] Error 1
make[1]: Leaving directory '/debcraft/source'
dh_auto_build: error: make -j4 INSTALL="install --strip-program=true" returned exit code 2
make: *** [debian/rules:20: binary] Error 2
dpkg-buildpackage: error: debian/rules binary subprocess failed with exit status 2
gbp:error: 'dpkg-buildpackage --post-clean --no-sign ' failed: it exited with 2
```

With the buildsystem defined for packages that are predominantly Go the build logs will have:

```
   dh_auto_build -O--buildsystem=golang
	cd obj-x86_64-linux-gnu && go version
```

The buildsystem autodetection does not seem to emit clearly anything about what it assumes the project is, or what alternatives it saw, or what users should do to manually pick the buildsystem. Thus I think it is more developer friendly to keep the buildsystem defined by default in all packages generated by a tool that is specifically designed to generate Go packaging.

jas4711:
Thanks for analysis!  Indeed presence of a Makefile seems to trigger non-go handling.  I agree that is a too fragile default, as many Go packages also ship a Makefile.  Maybe the automatic detection logic should be improved, and this change revisited, but until then: Closing the pull request...

PR #325, opened 2026-08-11
Append module version to package names

benignbala:
This change adds the version names to source and the -dev binary package when generating a new package configuration. This change gets the module information from just the argument passed to dh-make-golang


jas4711:
I didn't test the actual commit here, but I believe this is the right conceptually thing to do: using versioned Debian source package names avoids all the hassles with major API transitions.

make.go
```diff
@@ -104,9 +106,35 @@ func downloadFile(filename, url string) error {
...
+		return ""
+	}
+
+	suffix := strings.TrimPrefix(gopkg, repoRoot+"/")
+
+	if strings.Contains(suffix, "/") {
+		return ""
+	}
+
+	var versionSuffixRegExp = regexp.MustCompile(`^v[2-9][0-9]*$`)
```
toddy15: I think this is not completely correct. The regex does not match versions which start with a "1", e. g. from 10 to 19, like v10 or later v127, v1827 (if those get released ...).

So I think the correct regex should be this:
```go
regexp.MustCompile(`^v[0-9]+$`)
```

In order to not match v0 and v1, you need to check against those two versions specifically before checking with the rexex:

```go
if suffix == "v0" || suffix == "v1" {
	return ""
}
```

it would be good to include tests for this, see my second comment.

make_test.go
```diff
@@ -59,6 +58,8 @@ var nameFromGoPkg = []struct {
 	{"git.sr.ht/~sircmpwn/getopt", typeGuess, "", "golang-sourcehut-sircmpwn-getopt"},
 	{"golang.org/x/term", typeLibrary, "", "golang-golang-x-term"},
 	{"github.com/cli/cli", typeProgram, "gh", "gh"},
+	{"github.com/Debian/test-pkg/v2", typeLibrary, "", "golang-github-debian-test-pkg-v2"},
```
toddy15, line 60: Please add more tests for the version matching, see comment above. Possible tests:

```go
{"github.com/Debian/test-pkg/v0", typeLibrary, "", "golang-github-debian-test-pkg"},
{"github.com/Debian/test-pkg/v1", typeLibrary, "", "golang-github-debian-test-pkg"},
{"github.com/Debian/test-pkg/v15", typeLibrary, "", "golang-github-debian-test-pkg-v15"},
{"github.com/Debian/test-pkg/v37", typeLibrary, "", "golang-github-debian-test-pkg-v37"},
```

make.go
benignbala: Thanks @toddy15 - Sorry about that. It is a bad oversight from my side. Will fix that and add tests as suggested.

toddy15: Great, thanks! And one more nitpick: could you please run `go fmt` on your PR? Otherwise, the CI tests fail immediately.

benignbala:
The CI is failing in the deb install step. GitHub has had CI/CD outages regularly this week. Would it be ok to retrigger it please @toddy15 ? Thanks

toddy15:
@benignbala Sure, done now.

benignbala:
Thanks, @toddy15. The CI/CD checks are passing. Do I create a new PR for the other changes from the DebConf wiki page?

benignbala:
@toddy15  - What are the next steps in this? As I am new to this, I could be missing something and hence asking. Thanks

PR #328, opened 2026-08-28
Remove gitlab-ci comat file generation

auyer:
New packages are expected to use only the
'salsa-ci.yml' file, thus this should be
removed to avoid new packages using gitlab-ci.


auyer:
Hi @arthurbdiniz , please take a look at this.
I believe this will help avoiding this file in new packages.

ottok:
Are we sure all systems that create Go team package repos set the correct `debian/salsa-ci.yml` config path?

